PR #15, opened 2017-06-20
Added support of queue prefix configuration

### LINKED ISSUE #14: Job queues with custom prefix could not be discovered
The prefix of the job queue in bull is customizable according to the [reference doc](https://github.com/OptimalBits/bull/blob/master/REFERENCE.md#queue). But, with customized prefix,  those queues wouldn't be able to be listed in arena queues overview page. Probably the `options` [here](https://github.com/mixmaxhq/arena/blob/master/src/server/bull/index.js#L42) should be leveraged to allow carrying the prefix settings?

ccw:
To resolve #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,8 @@ Configure your queues in the "queues" key of [`index.json`](src/server/config/in
 }
 ```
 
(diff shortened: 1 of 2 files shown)


randallm:
Thanks!

PR #20, opened 2017-06-26
Small hotfixes/feature requests

randallm:


index.js
```diff
@@ -1,7 +1,9 @@
-const app = require('./src/server/app');
+const {app, server} = require('./src/server/app');
 
-module.exports = function(config) {
+module.exports = function(config, disableListen) {
```
chuym: What is this listener for?

If it's for conditionally stop the server for debugging/other development purposes, I'd suggest moving this to `src/server/app.js` and support passing a parameter like `node app.js --no-listen` or similar so that you don't have to close the server here

src/server/views/dashboard/templates/queueJobsByState.hbs
```diff
@@ -39,11 +39,9 @@
     </nav>
   </div>
   <div class="col-sm-6 text-right">
-    {{#eq state 'completed'}}
-      <button type="button" data-action="remove" data-queue-name="{{ queueName }}" data-queue-host="{{ queueHost }}" data-queue-state="{{ state }}" class="js-bulk-action btn btn-danger">
-        Remove Jobs
-      </button>
-    {{/eq}}
+    <button type="button" data-action="remove" data-queue-name="{{ queueName }}" data-queue-host="{{ queueHost }}" data-queue-state="{{ state }}" class="js-bulk-action btn btn-danger">
```
chuym: Does this action have a confirmation dialog? If not, it should have one.

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -2,20 +2,17 @@
   <h3>#{{ this.id }}</h3>
 {{/unless}}
 
+<h5>Actions</h5>
+<button class="btn btn-danger js-remove-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}" data-job-state="{{ jobState }}">
```
chuym: Similar to above, can't recall if we have a warning dialog for this

index.js
randallm: This is for the use case in which the user mounts Arena in [another express app as middleware](https://github.com/mixmaxhq/arena/issues/18). Thus, I think it may make more sense to support a way to programmatically toggle the listener rather than using a command line flag, since the user would presumably use Arena solely as a module. (fwiw, this is also how we use Arena in maintenance)

src/server/views/dashboard/templates/queueJobsByState.hbs
randallm: Yep!

Issue #29, opened 2017-07-15
When running as a node module, arena cannot be nested within a router


jrobson153:
I'm running arena as a node module alongside the rest of my app. For this use case it would be great to be able to nest it within a router so that the dashboard could be accessed at `someapp.com/arena/dashboard` for instance, like this:

```js
const arenaRouter = arena({ queues: getArenaConfig() }, true)
router.use('/arena', arenaRouter)
```

This is not currently possible, because all href links within the dashboard (and assets) are referenced directly at the root, i.e. `<a href="/dashboard/..."></a>`

randallm:
Added in v2.0.1.

sparebytes:
There another issue if arenaRouter is nested inside another router. EG: ```
const app = express();

const authenticatedRouter = express.Router();
app.use('/authenticated', authenticatedRouter);
// ...

const arenaRouter = arena({ ... }, { basePath: '/authenticated/arena', disableListen: true });
authenticatedRouter.use('/arena', arenaRouter);
```

In this example, the HTML anchors are correct `/authenticated/arena/...` but the server listens on `/authenticated/arena/authenticated/arena/...`. The only way to reconcile this is to use the arenaRouter from the app.

Issue #28, opened 2017-07-11
Next page link in pagination is broken


randallm:
https://github.com/mixmaxhq/arena/blob/master/src/server/views/dashboard/templates/queueJobsByState.hbs#L33

We need to unindent `queueHost` and `queueName`.

randallm:
Added in v2.0.1.

PR #30, opened 2017-07-15
Added baseURL to options when using as node module

jrobson153:
This is my solution to solve #29, which I just opened. You guys may want to implement it differently though, but for my use case this is working great.

With these changes you can initialize arena like this:
```js
const options = { 
    disableListen: true, 
    baseURL: '/arena'
}
const arenaRouter = arena({ queues: getQueueConfig() }, options)
router.use('/arena', arenaRouter)
```

The `baseURL` provided gets set as an option on express, and then is passed to each of the hbs templates to be prepended to each URL. It's also set as a data attribute on the body so that it can be accessed by the client js to prepend it when making api calls.

I've added backwards compatibility with the way `disableListen` is being handled currently.

I also noticed #28 while I was in there and fixed it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,11 @@
 const {app, server} = require('./src/server/app');
 
-module.exports = function(config, disableListen) {
+module.exports = function(config, options) {
+  options = typeof options == 'boolean' ? { disableListen: options } : options;
+  const { disableListen, baseURL } = options;
+
   app.set('bull config', config);
+  app.set('baseURL', (baseURL || ''));
 
   if (disableListen) server.close();
 
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -20,6 +20,8 @@ async function handler(req, res) {
   const endId = startId + pageSize - 1;
   const jobs = await queue[`get${_.capitalize(state)}`](startId, endId);
 
+  const baseURL = req.app.get('baseURL');
+
   let pages = _.range(page - 6, page + 7)
     .filter((page) => page >= 1);
   while (pages.length < 12) {
@@ -30,6 +32,7 @@ async function handler(req, res) {
   return res.render('dashboard/templates/queueJobsByState.hbs', {
     queueName,
     queueHost,
+    baseURL,
     state,
     jobs,
     jobsInStateCount: jobCounts[state],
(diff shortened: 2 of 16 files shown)


src/server/views/dashboard/queueJobsByState.js
```diff
@@ -30,6 +32,7 @@ async function handler(req, res) {
   return res.render('dashboard/templates/queueJobsByState.hbs', {
     queueName,
     queueHost,
+    baseURL,
```
wearhere: If you put this in [`res.locals`](http://expressjs.com/en/4x/api.html#res.locals) e.g. using middleware mounted before any routes, I think that will make it available to every template without having to be explicitly passed to each one.

index.js
```diff
@@ -1,7 +1,11 @@
 const {app, server} = require('./src/server/app');
 
-module.exports = function(config, disableListen) {
+module.exports = function(config, options) {
+  options = typeof options == 'boolean' ? { disableListen: options } : options;
+  const { disableListen, baseURL } = options;
```
wearhere: I think that the app can determine `baseUrl` itself using [`req.baseUrl`](http://expressjs.com/en/4x/api.html#req.baseUrl).

wearhere: And nit, wherever you reference the variable, can you capitalize it like `baseUrl` for consistency with that property?

src/server/views/dashboard/queueJobsByState.js
wearhere: It might also make `baseUrl` available to partials without you having to pass it through from the top level.

randallm:
Hey @jrobson153. Thanks for the PR! 😀 

We have an upcoming release lined up to fix this issue, which takes more or less the same approach and API as the one given in this PR (plus some style changes like the ones @wearhere mentioned above). 

I'll leave this PR open for visibility since we haven't cut the release yet — when that happens I'll ping you and close this thread.

jrobson153:
Thanks guys, and apologies for the delay! Sounds good, I'm happy to make those quick changes @wearhere mentioned if you still want me to, but it sounds like you guys already have it implemented. Let me know!

randallm:
@jrobson153, this has now made it into Arena v2.0.1, which is now published. If you have any thoughts or questions, please feel free to let us know!

jrobson153:
@randallm brilliant, thanks!

vjustov:
Was this ever really solved? i'm having the same issue. see the following repo for minimal reproduction: https://github.com/vjustov/arena-basePath-bug 

if you click in the link to a queue, it sends you to undefined/sendPriceAdjustmentEmail instead of queue/undefined/sendPriceAdjustmentEmail

PR #38, opened 2017-08-08
Disable caching, events for bee-queue

skeggse:
This can cause issues where stale data is displayed in the web interface, if the actual `Queue` instances have `sendEvents` disabled (or if Arena's `Queue` ends up out of sync somehow).


bradvogel:
thanks!

PR #46, opened 2017-08-30
Resolve #45 - progress bar display

### LINKED ISSUE #45: Job progress bar doesn't work with Bull v3
It appears that Bull v3 (precisely, 3.0.0-rc.4) returns a job's JSON object with `_progress` holding the progress field, instead of just `progress`. Thus, the progress bar in arena always effectively shows 0 percent.

logging the JSON object returned from the Queue class gives:
```
{ opts: { attempts: 1, delay: 0, timestamp: 1504061405201 },
  name: '__default__',
  queue: 
   Queue {
     name: 'disc ripping',
     token: '39700d85-c68a-427c-876a-e5aea5fb3d17',
     keyPrefix: 'bull',
     _initializing: 
      Promise {
        _bitField: 33554432,
        _fulfillmentHandler0: undefined,
        _rejectionHandler0: undefined,
        _promise0: undefined,
        _receiver0: undefined },
     handlers: {},
     processing: [],
     retrieving: 0,
     settings: 
      { lockDuration: 30000,
        stalledInterval: 30000,
        maxStalledCount: 1,
        guardInterval: 5000,
        retryProcessDelay: 5000,
        drainDelay: 2000,
        lockRenewTime: 15000 },
     _events: { error: [Function] },
     _eventsCount: 1,
     timers: TimerManager { idle: true, listeners: [], timers: {} },
     moveUnlockedJobsToWait: [Function: bound ],
     processJob: [Function: bound ],
     getJobFromId: [Function: bound ],
     IS_BEE: false },
  data: { something: 'somethingelse' },
  _progress: 100,
  delay: 0,
  timestamp: 1504061405201,
  stacktrace: [],
  returnvalue: 18,
  attemptsMade: 0,
  id: '20',
  finishedOn: 1504061415220,
  processedOn: 1504061405209,
  failedReason: undefined,
  jobState: 'completed',
  queueHost: 'ncodr',
  queueName: 'disc ripping',
  basePath: '' } { name: 'log',
  hash: {},
  data: 
   { exphbs: 
      { cache: false,
        view: 'dashboard/templates/jobDetails',
        layout: '/ncodr/node_modules/bull-arena/src/server/views/layout',
        data: undefined,
        helpers: {},
        partials: [Object],
        filePath: '/ncodr/node_modules/bull-arena/src/server/views/dashboard/templates/jobDetails.hbs' },
     _parent: { exphbs: [Object] },
     root: 
      { settings: [Object],
        Queues: [Object],
        basePath: '',
        queueName: 'disc ripping',
        queueHost: 'ncodr',
        jobState: 'completed',
        job: [Object],
        _locals: {},
        cache: false } } }
```

Which ends up with a display looking like:
image

instead of a full across progress bar. In my  troubleshooting, just changing to `this._progress` in `src/server/views/partials/dashboard/jobDetails.hbs` solves it. However, I would guess that Bee uses `progress`, and so the difference would need to be accounted for in arena somehow.
Maybe it's possible to alias `_progress` to `progress` if bull is being used from within `jobDetails.js`?

However, interestingly, when adding the following line in `jobDetails.js` (around line 28), I get correct output in the log:
`  console.log('jobdetails -- progress: ' + job.progress());`

aztechian:
Fix logic for displaying progress bar section of job detail page.
Don't display the progress bar section at all if the Queue type is Bee. If the queue is not Bee (ie, it's Bull) then display the progress display, but use it as `_progress` instead of `progress`

### DIFF
--- a/src/server/views/partials/dashboard/jobDetails.hbs
+++ b/src/server/views/partials/dashboard/jobDetails.hbs
@@ -45,6 +45,7 @@
   </div>
 </div>
 
+{{#unless this.queue.IS_BEE}}
 <h5>Progress</h5>
 <div class="progress">
   <div class="progress-bar
@@ -52,13 +53,14 @@
               progress-bar-danger
               {{/eq}}"
        role="progressbar"
-       aria-valuenow="{{ this.progress }}"
+       aria-valuenow="{{ this._progress }}"
        aria-valuemin="0"
        aria-valuemax="100"
-       style="width: {{ this.progress }}%; min-width: 2em;">
-    {{ this.progress }}%
+       style="width: {{ this._progress }}%; min-width: 2em;">
+    {{ this._progress }}%
   </div>
 </div>
+{{/unless}}
 
 {{#if this.returnvalue}}
 <pre>{{ this.returnvalue }}</pre>


bradvogel:
Thanks - released as 2.1.3

PR #47, opened 2017-08-30
Support bulk actions on jobs with encoded characters

eddieloeffen:
The UI sends an array of url encoded Job Ids when doing things like Bulk Remove Jobs. 

This change allows the server to decode these ids to resolve to the correct job id.

### DIFF
--- a/src/server/views/api/bulkAction.js
+++ b/src/server/views/api/bulkAction.js
@@ -21,7 +21,7 @@ function bulkAction(action) {
 
     try {
       if (!_.isEmpty(jobs)) {
-        const jobsPromises = jobs.map((id) => queue.getJob(id));
+        const jobsPromises = jobs.map((id) => queue.getJob(decodeURIComponent(id)));
         const fetchedJobs = await Promise.all(jobsPromises);
 
         const actionPromises = fetchedJobs.map(job => job[action]());


src/server/views/api/bulkAction.js
```diff
@@ -21,7 +21,7 @@ function bulkAction(action) {
 
     try {
       if (!_.isEmpty(jobs)) {
-        const jobsPromises = jobs.map((id) => queue.getJob(id));
+        const jobsPromises = jobs.map((id) => queue.getJob(decodeURIComponent(id)));
```
bradvogel: shouldn't these be decoded automatically by Express?

eddieloeffen: This is the JSON string that is sent by the UI:

`{"queueName":"test concurrent queue","action":"remove","jobs":["repeat%3Atest%20task%3A%3A1504132820000"]}`

Which is parser by the express JSON body parser.

I cannot see any options to the JSON body parser to call decodeURIComponent:
https://github.com/expressjs/body-parser#bodyparserjsonoptions

As far as I can tell, express is just parsing the JSON as it is sent?

bradvogel: Oh I see. Ok, this works

PR #56, opened 2017-09-30
fix hostId/host misuse

slavaGanzin:
> The name, port, host, and hostId fields are required. hostId can be given any name, so it is recommended to give it a helpful name for reference.

As I understand hostId is alias for host, so it should be used only in templates...

p.s. Why host/port have no default value? 127.0.0.1:6379 would fit for newbies (like me)


slavaGanzin:
It worked without my PR. Ah...

PR #78, opened 2017-12-29
updated bull to 3.3.7

mathieupythian:
This fixes a crash of the Arena UI when the job completes.


bradvogel:
thanks. Published as 2.3.0

PR #80, opened 2018-01-04
Styling and formatting fixes for jobDetails view

emhagman:
1. Added `<pre>` formatting using `highlight.js`
2. Fixed an issue with the `<pre>` tag and whitespace control when viewing the template inline
3. Set a max height of 250px to the `<pre>` tag with overflow-y: scroll
4. Put the "Permalinks" on a new row so that long id's don't overflow outside of tag (we used uuid.v4() for instance, a pretty long id)
5. Formatted `this.returnvalue` using the json helper as well in case the returnvalue is not just a string

If you don't want to accept all changes, I can make a separate pull request for just the whitespace control so that the `<pre>` tags look proper on the inline view.

### DIFF
--- a/src/server/views/dashboard/templates/queueJobsByState.hbs
+++ b/src/server/views/dashboard/templates/queueJobsByState.hbs
@@ -101,7 +101,7 @@
               </h4>
             </a>
             <div id="collapse{{encodeIdAttr this.id}}" class="collapse">
-              {{> dashboard/jobDetails this basePath=../basePath displayJobInline=true queueName=../queueName queueHost=../queueHost jobState=../state }}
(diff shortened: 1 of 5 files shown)


src/server/views/dashboard/templates/queueJobsByState.hbs
```diff
@@ -101,7 +101,7 @@
               </h4>
             </a>
             <div id="collapse{{encodeIdAttr this.id}}" class="collapse">
-              {{> dashboard/jobDetails this basePath=../basePath displayJobInline=true queueName=../queueName queueHost=../queueHost jobState=../state }}
+              {{~> dashboard/jobDetails this basePath=../basePath displayJobInline=true queueName=../queueName queueHost=../queueHost jobState=../state }}
```
wearhere: I did not know about using tildes to control whitespace, neat!

bradvogel:
Thanks! published as 2.3.1

PR #81, opened 2018-01-04
Custom Redis connection; documentation

### LINKED ISSUE #67: Need to pass skipVersionCheck redis option to bull.
Currently there seems to be no ability to pass this flag in the redis options.

vhf:
This PR fixes #67. 

It also fixes a bug related to Bull configuration. Creating a Bull queue using a URL as configuration is done by passing it as a string 2nd argument, not as `{url}` object 2nd argument: https://github.com/OptimalBits/bull#basic-usage

I also enhanced the documentation a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,31 +19,94 @@ For a quick introduction to the motivations for creating Arena, read *[Interacti
 
 #### Prerequisites
 
-Configure your queues in the "queues" key of [`index.json`](src/server/config/index.json). Queues take the following format:
+Configure your queues in the "queues" key of [`index.json`](src/server/config/index.json).
 
-```json
+Queues are JSON objects. Here are the configuration keys that are common to all three configuration ways:
+
+```js
 {
+  // required string
   "name": "my_queue",
-  "port": 6381,
-  "host": "127.0.0.1",
-  "hostId": "AWS Server 2"
+
+  // host display name, give it a helpful name for reference
+  // required string
+  "hostId": "Queue Server 1",
+
+  // optional string
+  // default: null (will assume Bull)
+  "type": "bee",
+
+  // queue keys prefix
+  // optional string
+  // default: "bq" for Bee, "bull" for Bull
+  "prefix": "foo"
 }
 ```
 
-The `name`, `port`, `host`, and `hostId` fields are required. `hostId` can be given any name, so it is recommended to give it a helpful name for reference. 
-Optionally, you can also pass in `db` and `password` to configure redis credentials, or `prefix` to specify the customized prefix of the queue.
+The required `name` and `hostId` have to be present in any of the following JSON objects, the optional keys can be present in them.
 
-You can also provide a `url` field instead of `host`, `port` `db` and `password`.
+The three ways in which you can configure the client are:
 
-To specify a custom file location, see "Running Arena as a node module".
+##### 1. port/host
 
-*Note that if you happen to use Amazon Web Services' Elasticache as your Redis host, check out http://mixmax.com/blog/bull-queue-aws-autodiscovery*
+```js
+{
+  // hostname or IP
+  // required string
+  "host": "127.0.0.1",
+
+  // optional number
+  // default: 6379
+  "port": 6379,
+
+  // optional string
+  "password": "hello",
+
+  // optional number
+  // default: 0
+  "db": 1,
+}
+```
+
+##### 2. URL
+
+You can also provide a `url` field instead of `host`, `port`, `db` and `password`.
+
+```js
+{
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,20 +30,64 @@ Configure your queues in the "queues" key of [`index.json`](src/server/config/in
 }
```
wearhere: Now that there are several ways of configuring the Redis client, I think we should make it clearer up top here that `name` is the only always-required parameter. Perhaps replacing "Queues take the following format:" to the end of this section with:

---
Queues are JSON objects. The only required key is "name", the name of the queue. The other keys describe how to configure the underlying Redis client, and vary. Here is a queue that initializes the client to connect to a specific host and port:

```json
{
  "name": "my_queue",
  "port": 6381,
  "host": "127.0.0.1",
  "hostId": "AWS Server 2"
}
```

The ways in which you can configure the client are:

README.md
```diff
@@ -30,20 +30,64 @@ Configure your queues in the "queues" key of [`index.json`](src/server/config/in
...
 ```
 
-The `name`, `port`, `host`, and `hostId` fields are required. `hostId` can be given any name, so it is recommended to give it a helpful name for reference. 
+There are several ways of configuring a queue:
+
+##### port host [db] [password]
+
+The `name`, `port`, `host`, and `hostId` fields are required. `hostId` can be given any name, so it is recommended to give it a helpful name for reference.
 Optionally, you can also pass in `db` and `password` to configure redis credentials, or `prefix` to specify the customized prefix of the queue.
```
wearhere: Can you show all of these keys in the JSON blob, with example values and per-option documentation saying whether the option is required / optional / has a default? Perhaps like https://github.com/rollup/rollup-plugin-node-resolve#usage.

wearhere: Then you won't need these paragraphs alongside the code sample.

README.md
```diff
@@ -30,20 +30,64 @@ Configure your queues in the "queues" key of [`index.json`](src/server/config/in
...
+  "name": "my_queue",
+  "port": 6381,
+  "host": "127.0.0.1",
+  "hostId": "AWS Server 2"
+}
+```
+
+##### URL
+
 You can also provide a `url` field instead of `host`, `port` `db` and `password`.
```
wearhere: It looks like `hostId` is also not necessary, in this scenario? I would avoid repeating the keys from previous sections, but rather make it clear that each section shows a complete queue object.

README.md
```diff
@@ -30,20 +30,64 @@ Configure your queues in the "queues" key of [`index.json`](src/server/config/in
...
+Arena is compatible with both Bee and Bull.
+If you need to pass some specific configuration options directly to the redis client library your queue uses, you can also do so.
+
+Bee uses node [redis](https://www.npmjs.com/package/redis) client, Bull uses [ioredis](https://www.npmjs.com/package/ioredis) client.
+These clients expect different configurations options.
+
+```json
+{
+  "name": "my_queue",
+  "prefix": "q",
```
wearhere: Doesn't look like this is documented in this section. Is this a key that applies independent of how the Redis client is configured? If so, please document it up top where `name` is talked about and maybe don't repeat it in the Redis configuration examples, unless it's a required key.

src/server/queue/index.js
```diff
@@ -29,26 +29,28 @@ class Queues {
...
-      prefix,
-      redis: url || { port, host, db, password }
-    };
-
     let queue;
     if (isBee) {
-      Object.assign(options, {
+      const options = {
+        prefix,
+        redis: url || redis || { port, host, db, password },
```
wearhere: nit: even though we don't expect multiple of the conditions here to be truthy, I think that `redis` should be ordered first since that's the most powerful option. If that happened to be an existing client, I especially don't think the developer would want another client to be created.

README.md
vhf: I liked your suggestion but after taking into account your other comments I went another way: I grouped the common config keys in here, numbered the three config ways and avoided repeated what is common to all of them. I think it's better that way, let me know what you think!

README.md
vhf: Done.

README.md
vhf: Done.

README.md
vhf: Moved to the 'common' section.

src/server/queue/index.js
vhf: Good catch, yep!

README.md
wearhere: Great way to solve it!

wearhere:
Published as 2.4.0.

vhf:
Thanks a lot!

PR #88, opened 2018-02-13
Use node:alpine image for docker

yasinuslu:
Currently docker image is building over `node:8.1.2`, it uses debian jessie and it's too big. Arena only needs a node environment so there seems to be no reason not to use `node:alpine`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.1.2
+FROM node:alpine
 
 EXPOSE 4567
 


ttacon:
Thanks for the PR! While I don't disagree with the sentiment, can you post the built image size differences before we merge this?

yasinuslu:
Of course, here is a screenshot i took in my machine:
image

pjanuario:
I was thinking about the same, I don't see any particular reason to not use alpine base image.If not it could provide both flavors but honestly I think is too much effort to make both availble and I don't see much benefits.

ttacon:
We now use Alpine, and have for a little while - closing - thanks for the PR!

Issue #90, opened 2018-03-13
Replace dependencies on CDN with local ones


marcbachmann:
I'm working on a project where we there's no decent internet connection setup.
This basically renders the dashboard useless.
What's your opinion regarding replacing the assets hosted on cdn.jsdelivr.net with local ones.
This will affect the jquery, bootstrap and highlightjs.

https://github.com/bee-queue/arena/blob/be866d368b44f95dbe4424ba71d6b5ec7d7843e9/src/server/views/layout.hbs#L9-L16

ttacon:
I'd prefer we don't do a full swap out here (as we'd lose the benefit of the CDN for other users 😄 ). However, I wouldn't be strongly opposed to make the location a template variable that defaults the the CDN host but that could be overridden by a configuration value.

marcbachmann:
https://github.com/bee-queue/arena/pull/91 👏

PR #91, opened 2018-03-25
Added bundled js and css file

hoo29:
First attempt for a fix for https://github.com/bee-queue/arena/issues/90

To me useCdn seems more like a listenerOption than queue configuration which is why I put it there. I have not used handlebarjs before so I imagine you'll have a few comments!


hoo29:
@ttacon - do you or someone else have time to have a look at this?

ttacon:
I'll try to take a look before the end of the week!

index.js
```diff
@@ -7,6 +7,7 @@ function run(config, listenOpts = {}) {
   const {app, Queues} = Arena();
 
   if (config) Queues.setConfig(config);
+  if (typeof listenOpts.useCdn !== 'undefined') Queues.setUseCdn(listenOpts.useCdn);
```
ttacon: This doesn't seem like it's defaulting to `false` as per [here](https://github.com/bee-queue/arena/pull/91/files#diff-04c6e90faac2675aa89e2176d2eec7d8R157)? I see that you're doing it [here](https://github.com/bee-queue/arena/pull/91/files#diff-623167d1c54ecd199ee8a745a96c568fR9), but we should do it inside the app initialization here instead of pushing it down to the queues.

src/server/queue/index.js
```diff
@@ -18,6 +19,14 @@ class Queues {
     this._config = config;
   }
 
+  setUseCdn(useCdn) {
```
ttacon: Any reason not use [getters and setters here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)?

src/server/views/helpers/handlebars.js
```diff
@@ -41,11 +41,15 @@ const helpers = {
 
   encodeIdAttr(id) {
     return id.replace(/:| /g, '');
-  }
+  },
 };
 
-module.exports = function registerHelpers(hbs) {
+module.exports = function registerHelpers(hbs, queues) {
```
ttacon: Instead of passing in a reference to the queues itself, let's make this an options argument and destructure the `queues` inside of it.

hoo29:
Thanks for the comments, hopefully I have addressed them all with the latest commit

src/server/views/helpers/handlebars.js
```diff
@@ -41,11 +41,15 @@ const helpers = {
 
   encodeIdAttr(id) {
     return id.replace(/:| /g, '');
-  }
+  },
```
ttacon: nit: remove the `,`

src/server/views/layout.hbs
```diff
@@ -1,54 +1,69 @@
 <html lang="en">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <title>Arena</title>
+<head>
```
ttacon: nit, can you fix the indentation here and throughout? `<head>` should be indented as it's nested inside `<html>`

hoo29:
@ttacon - hopefully now!

src/server/views/layout.hbs
```diff
@@ -1,26 +1,34 @@
...
 
+    {{#if (useCdn)}}
     <!-- Bootstrap core CSS -->
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/bootstrap/3.3.7/css/bootstrap.min.css">
-    <link rel="stylesheet" href="{{ basePath }}/dashboard.css">
-
     <!-- JSON Viewer -->
     <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/default.min.css">
-    <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script> {{else}}
```
ttacon: Let's move the `{{else}}` and the closing `{{/if}}` to their own lines for readability.

src/server/views/layout.hbs
```diff
@@ -46,9 +54,11 @@
       </div>
     </div>
 
-    <input type="hidden" id="basePath" value="{{ basePath }}">
-
-    <script type="text/javascript" src="//cdn.jsdelivr.net/g/jquery@3.2.1,bootstrap@3.3.7"></script>
+    <input type="hidden" id="basePath" value="{{ basePath }}"> {{#if (useCdn)}}
```
ttacon: Same as above, let's put the markup control logic statements on their own lines for readability.

Issue #92, opened 2018-03-26
Readme Improvements


tim-soft:
I had a real heck of a time understanding how to build a working config for arena. I think it would be a lot more helpful to show a full working config for bee-queue and bull instead of what is in the readme currently. 

Here's a working outline for a config for bull setup as an express middleware. I hope this helps other people get up and running!

```javascript
import Arena from 'bull-arena';

const arenaConfig = Arena({
  queues: [
    {
      // Name of the bull queue, this name must match up exactly with what you've defined in bull
      name: "Notification_Emailer",

      // host name or queue prefix, you can put whatever you want
      hostId: "MyAwesomeQueues",

      // Redis auth
      redis: {
        port: /* Your redis port */,
        host: /* Your redis host domain*/,
        password: /* Your redis password */,
      },
    },
  ],
},
{
  // Make the arena dashboard become available at {my-site.com}/arena
  basePath: '/arena',

  // Let express handle the listening
  disableListen: true
});

// Make arena's resources (js/css deps) available at the base app route
app.use('/', arenaConfig);
```

aurelben:
just what I need ;) 

Thanks a lot

bradvogel:
Thanks so much for filing this! Would you mind making a quick PR to update the readme?

minheq:
omg thanks!

ttacon:
This has been added to the README [here](https://github.com/bee-queue/arena#example-config-for-bull) and references @tim-soft as the author. Thanks again @tim-soft!

PR #94, opened 2018-05-04
Add example config from issue #92 to the README.

ttacon:


ttacon:
Adding the example config from #92 and give credit for it to @tim-soft.

PR #96, opened 2018-05-14
Ensure job progress is displayed (compatibility with latest version of bull)

mattbrictson:
As of bull 3.4.2 (and probably earlier), the progress of a job is provided in `progress`, not `_progress`. This commit updates the view template to match.

This fixes a problem where the progress of any bull job would always appear in arena as "0%". Now the correct progress is shown.

See: https://github.com/OptimalBits/bull/blob/v3.4.2/lib/job.js#L129


bradvogel:
Published as bull-arena@2.4.3 - thanks!

mattbrictson:
🙌

PR #103, opened 2018-06-13
Fix bull job progress

### LINKED ISSUE #102: Job progress is broken
Job progress is always 0%. 

I believe [this](https://github.com/bee-queue/arena/commit/33f730a69f45c2887f6fcd1ad48e42250b214782#diff-9a220580a657ea11601817f0fa5303f9) commit introduced the problem.

Does anyone else face the same issue? If it's not just me, maybe we should revert that commit?

dostu:
Fixes #102.

### DIFF
--- a/src/server/views/partials/dashboard/jobDetails.hbs
+++ b/src/server/views/partials/dashboard/jobDetails.hbs
@@ -54,11 +54,11 @@
               progress-bar-danger
               {{/eq}}"
        role="progressbar"
-       aria-valuenow="{{ this.progress }}"
+       aria-valuenow="{{ this._progress }}"
        aria-valuemin="0"
        aria-valuemax="100"
-       style="width: {{ this.progress }}%; min-width: 2em;">
-    {{ this.progress }}%
+       style="width: {{ this._progress }}%; min-width: 2em;">
+    {{ this._progress }}%
   </div>
 </div>
 {{/unless}}


skeggse:
Per 33f730a, different versions of bull store the progress in different fields. Can you make this support both fields?

istvanp:
I get the same issue on both latest versions of bull and arena. From what I can tell, if you blame the file, that progress variable has not changed in 5 years and the `Job.prototype.progress` wasn't changed recently: https://github.com/OptimalBits/bull/blame/1c01719232e954a37ac362ff062e75c6422488a5/lib/job.js 
So I'm confused why https://github.com/bee-queue/arena/commit/33f730a69f45c2887f6fcd1ad48e42250b214782 was done (and is working?).

Looking closer at the code in Bull, using `this.progress` will return the function itself, not the value which is why it's not updating. To return the value, you need to use `this.progress()` or `this._progress` as per this PR (the former is better IMO though).

dostu:
`Job.prototype.progress()` is used to update the progress so we can't use it here.

istvanp:
@dostu Without an argument it returns the current progress value[1].
[1] https://github.com/OptimalBits/bull/blob/1c01719232e954a37ac362ff062e75c6422488a5/lib/job.js#L108

milesalex:
Excited for this!

milesalex:
Any update here?

ttacon:
Thanks for the PR!

Issue #105, opened 2018-07-02
`app.use('/jobs', arena)` + `basePath : '/jobs'` listens on "/jobs/jobs" instead of "/jobs"


Ilshidur:
As said in the title, I think the routing is incorrect when using this module with nested routes.

Example :

```javascript
const arena = Arena({
  queues: [
    {
      // ...
    }
  ],
}, {
  basePath: '/jobs',
  disableListen: true,
});

// [...]

app.use('/jobs', arena);
// Only reachable by accessing `/jobs/jobs` route
```

When accessing the dashboard on `/jobs/jobs`, the assets can't be accessed and the assets requests results in a 404 error : `/jobs/dashboard.css` (happens with the other assets).

It this an intended behavior ? If so, I didn't find any doc about this :-(
Is this a bug ? If so, [I opened a MR solving this problem](https://github.com/bee-queue/arena/pull/106) in case this is indeed a bug.

Thanks for this great lib by the way !

hraban:
Ran into the same problem. Work around:

```ts
const app = express();
// ...

// URL Rewriting rigmarole necessary because bull-arena expects to be mounted
// on root path. I don't want to do that here because I want to encapsulate
// the bull arena, most importantly for security reasons (password and CSP).
const arena = express.Router();
arena.use((req, res, next) => { req.url = `/arena${req.url}`; next(); });
arena.use(new Arena({queues}, {basePath: '/arena', disableListen: true, ...}));
app.use('/arena', arena);
```

PR #106, opened 2018-07-02
fix: base path with nested routes

Ilshidur:
Fixes https://github.com/bee-queue/arena/issues/105 if this is confirmed as a bug.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,8 +11,8 @@ function run(config, listenOpts = {}) {
 
   app.locals.basePath = listenOpts.basePath || app.locals.basePath;
 
-  app.use(app.locals.basePath, express.static(path.join(__dirname, 'public')));
-  app.use(app.locals.basePath, routes);
+  app.use('/', express.static(path.join(__dirname, 'public')));
+  app.use('/', routes);
 
   const port = listenOpts.port || 4567;
   if (!listenOpts.disableListen) {


pbassut:
Can we get this in?

index.js
```diff
@@ -11,8 +11,8 @@ function run(config, listenOpts = {}) {
 
   app.locals.basePath = listenOpts.basePath || app.locals.basePath;
 
-  app.use(app.locals.basePath, express.static(path.join(__dirname, 'public')));
-  app.use(app.locals.basePath, routes);
+  app.use('/', express.static(path.join(__dirname, 'public')));
```
chuym: Näive question: If `app.locals.basePath` is set to `/` through `listenOpts.basePath`, will this achieve the same goal?

Ilshidur: If I understand well :

Setting the `basePath` to `'/'` when using the middleware on a `/jobs` route would set the base route of all the `<a>` links (and `<link>`) to `/`. This will make the stylesheets queries fail because the browser will request on non existing routes, based on `/` instead of `/jobs`.

Example below :

https://github.com/bee-queue/arena/blob/eef2b1a0c8ccb7a202260cd6ca419195b28c60fc/src/server/views/layout.hbs#L18

<hr/>

You can find a concrete use case in the [issue related to this PR](https://github.com/bee-queue/arena/issues/105) where the basePath would be `/jobs` and the route would end up being on `/jobs/jobs` instead of `/jobs`.

I believe changing `app.locals.basePath` in the source code to `'/'` would prevent the middleware's routes to be set on a doubled base path 😄

chuym: Thank you for the explanation!

I see the issue here, if we go with this fix route it means that `Arena` needs to be initialized with `basePath` set to the same route used in the express route mounting arena. That is, if we have:

```
express.route('/jobs', arena)
```

then we have to initialize arena like so:

```
const arena = new Arena(opts, {
  basePath: '/jobs'
})
```

so that assets can be retrieved, because we use `basePath` when fetching assets. 

There's another proposed fix here: https://github.com/bee-queue/arena/pull/119 which deals with this by using `req.baseUrl` to load assets, which will use the correct path all the time without the need to pass a `basePath` that matches the parent route. Could you take a quick glance to check if it fixes the issue you're seeing? Note that it adds some other stuff that I've asked to remove so we can keep it as a fix for #105 

With that said, if we end up merging #119, the `basePath` option becomes pretty much useless because users can mount arena to any route they want, since it's just a middleware. #119 adds a way to resolve local assets with the right path by using `req.baseUrl`.

Ilshidur: I saw the other PR just after posting my message ! 😄 

I think the PR https://github.com/bee-queue/arena/pull/119 seems more appropriate, because the user of the lib would not have to set a `basePath` option, thus repeating themselves by also declaring the base URL in the mounted route. And - as you said - `basePath` would become useless.

My PR was just a basic workaround that I suggested in a hurry. I suggest to close it and focus on https://github.com/bee-queue/arena/pull/119 later on.

ttacon:
Closing as #119 was merged. Thanks for the PR and discussion!

PR #109, opened 2018-07-20
Add health check for kubernetes

milesalex:
I'd like to use the image that y'all host but with a little health check for a k8s deployment


ttacon:
Hey! Thanks for the PR! Any reason not to use the [module functionality](https://github.com/bee-queue/arena#running-arena-as-a-node-module)?

milesalex:
@ttacon It seemed more simple to me to use the pre-packaged image with the github repo backing it with a health check included so I don't need to repeatedly pull new updates into something I host.

Does that make sense? I totally get if you think this is out of scope for arena.

milesalex:
Agree! What would you like to see and how far off is what I have?

milesalex:
Using root for now.

PR #114, opened 2018-07-26
add host opt

dotob:
add listenOpts host for selecting ip of host to listen on

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,8 +15,9 @@ function run(config, listenOpts = {}) {
   app.use(app.locals.basePath, routes);
 
   const port = listenOpts.port || 4567;
+  const host= listenOpts.host || '0.0.0.0'; // default: listen to all ips
   if (!listenOpts.disableListen) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -15,8 +15,9 @@ function run(config, listenOpts = {}) {
   app.use(app.locals.basePath, routes);
 
   const port = listenOpts.port || 4567;
+  const host= listenOpts.host || '0.0.0.0'; // default: listen to all ips
```
ttacon: One tiny nit, since the reason this will seem like a different IP is because we're binding to a different network interface, let's say that, so let's have the comment say:

`// Default: listen to all network interfaces.`

dotob:
@ttacon better?

ttacon:
Yup! I'll merge this now. My one concern is that this no longer supports IPv6 out of the box ([see default network interface binding here](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback)). We can handle that when necessary, though.

ttacon:
Thanks again for the PR!

PR #119, opened 2018-09-07
fix: base path with nested routes which may have wildcards

### LINKED ISSUE #105: `app.use('/jobs', arena)` + `basePath : '/jobs'` listens on "/jobs/jobs" instead of "/jobs"

ronnievdv:
Fixes #105 but also allows for the route to have a wildcard

```
const arena = Arena({
  queues: [
    {
      // ...
    }
  ],
}, {
  basePath: '/jobs',
  disableListen: true,
});

// [...]

app.use('/user/:id', arena);
```

Will result in a path /user/:id/jobs which can then be matched to an actual url like user/1/jobs

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,10 +9,10 @@ function run(config, listenOpts = {}) {
   if (config) Queues.setConfig(config);
   Queues.useCdn = typeof listenOpts.useCdn !== 'undefined' ? listenOpts.useCdn : true;
 
-  app.locals.basePath = listenOpts.basePath || app.locals.basePath;
+  app.locals.appBasePath = listenOpts.basePath || app.locals.appBasePath;
 
-  app.use(app.locals.basePath, express.static(path.join(__dirname, 'public')));
-  app.use(app.locals.basePath, routes);
+  app.use(app.locals.appBasePath, express.static(path.join(__dirname, 'public')));
+  app.use(app.locals.appBasePath, routes);
 
   const port = listenOpts.port || 4567;
   const host= listenOpts.host || '0.0.0.0'; // Default: listen to all network interfaces.
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -22,7 +22,7 @@ module.exports = function() {
   const queues = new Queues(defaultConfig);
   require('./views/helpers/handlebars')(handlebars, { queues });
   app.locals.Queues = queues;
-  app.locals.basePath = '';
+  app.locals.appBasePath = '';
 
   app.set('views', `${__dirname}/views`);
   app.set('view engine', 'hbs');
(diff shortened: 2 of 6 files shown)


src/server/views/layout.hbs
```diff
@@ -12,16 +12,16 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/bootstrap/3.3.7/css/bootstrap.min.css">
     <!-- JSON Viewer -->
     <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/default.min.css">
-    <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js" nonce="{{ nonce }}"></script>
```
chuym: Could you elaborate a bit more on the purpose of this `nonce`? Is there a warning or similar that requires the use of this?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "bull-arena",
+  "name": "@sortdinc/bull-arena",
```
chuym: Please revert this change in the PR

package.json
```diff
@@ -23,5 +23,5 @@
     "node": ">=7.6.0"
   },
   "repository": "https://github.com/bee-queue/arena.git",
-  "version": "2.4.5"
+  "version": "2.4.6"
```
chuym: Please revert this change in the PR

src/server/views/layout.hbs
chuym: If you would like to contribute this, please do so in a separate PR with the motivation behind this change. Let's keep this only about #105

ronnievdv:
@chuym  I did a rebase to remove the last two commits, I'll add a separate pull request for adding the nonce.

chuym:
Thank you so much for your fix @ronnievdv ! this is a very wanted fix, will merge now.

PR #122, opened 2018-09-20
Support Redis-over-TLS

timcosta:
Currently Arena does not pass forward any TLS opts it receives as part of the Redis config. If a TLS opt is supplied, it should be passed to the queue Redis configuration so that Arena can read from TLS-enabled Redis instances.

### DIFF
--- a/src/server/queue/index.js
+++ b/src/server/queue/index.js
@@ -39,12 +39,13 @@ class Queues {
       return this._queues[queueHost][queueName];
     }
 
-    const { type, name, port, host, db, password, prefix, url, redis } = queueConfig;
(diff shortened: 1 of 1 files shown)


timcosta:
Any chance of this getting merged and published soon? We're going to need to explore other options unfortunately if not.

ttacon:
We'll be taking a look soon, thanks for the contribution!

Out of my own personal curiosity, how's the performance of the TLS handling via node vs using something like `stunnel` to offload via a private local port?

ttacon:
Merged and published as 2.5.0.

timcosta:
Honestly we haven’t had any issues using TLS in node. We saw no measurable performance decreases when we added TLS to our caching and job queue clusters. 

Thank you for publishing!!

PR #126, opened 2018-10-24
Fixes issue preventing expand/collapse functionality from working when job ids contain special characters

### LINKED ISSUE #116: Custom Job id and collapse error
If `/` or other special character (not escaped in the `encodeIdAttr` function) are present in the job id, the job details when collapsed won't open and an error is thrown in the console by `jquery.bootstrap`.

This is a problem when using custom job ids, and in my case with those special characters on it.

To fix this you can simply replace in file [src/server/views/helpers/handlebars.js:43](https://github.com/bee-queue/arena/blob/master/src/server/views/helpers/handlebars.js#L43) the code:
`return id.replace(/:| /g, '');`
with
`return Buffer.from(id).toString('base64').replace(/=/g, '');`
or any other hash, `encodeURIComponent` doesn't work either as `%` also throws an error.

Or simply use a sequential id, really.

gcox:
Fixes #116 

I did try escaping these characters first but it did not seem to help at all, so just went with a hash since the result is not currently being used for anything other than generating a unique attribute value.

### DIFF
--- a/src/server/views/helpers/handlebars.js
+++ b/src/server/views/helpers/handlebars.js
@@ -41,7 +41,7 @@ const helpers = {
   },
 
   hashIdAttr(id) {
-    return crypto.createHash('md5').update(id).digest('hex');
+    return crypto.createHash('sha256').update(id).digest('hex');
   }
 };
 
--- a/src/server/views/helpers/handlebars.js
+++ b/src/server/views/helpers/handlebars.js
@@ -1,3 +1,4 @@
+const crypto = require('crypto')
 const _ = require('lodash');
 const Handlebars = require('handlebars');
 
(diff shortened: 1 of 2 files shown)


src/server/views/helpers/handlebars.js
```diff
@@ -39,8 +40,8 @@ const helpers = {
     block.push(options.fn(this));
   },
 
-  encodeIdAttr(id) {
-    return id.replace(/:| /g, '');
+  hashIdAttr(id) {
+    return crypto.createHash('md5').update(id).digest('hex');
```
ttacon: Any reason not to use `sha256` instead of `md5`?

gcox: Not a good enough reason. Let me know Github's suggestion syntax doesn't work for you and I'll actually commit/push the change.

```suggestion
    return crypto.createHash('sha256').update(id).digest('hex');
```

ttacon:
Cheers!

PR #133, opened 2018-11-13
support "paused" as a state for bull queues. closes #132

### LINKED ISSUE #132: Paused jobs not properly handled by queue state url handler
Because `paused` is not in the [accepted list of bull queue states](https://github.com/bee-queue/arena/blob/master/src/server/views/helpers/queueHelpers.js#L27), it is treated as [an individual job](https://github.com/bee-queue/arena/blob/cbfa82ac09c12ae074a5fe8c23d52c64ee178de2/src/server/views/dashboard/index.js#L10-L11) which fails on `hgetall` with a `WRONGTYPE` error from Redis:

> WRONGTYPE Operation against a key holding the wrong kind of value

mshwery:
Previously the `bull-arena` package did not support paused as a valid bull queue state, despite the UI showing it as one. This ended up causing server crashes due when you navigate to the `paused` queue state due to trying to read the key from redis as an individual job, rather than a list (HGETALL vs LRANGE).

screen shot 2018-11-13 at 1 30 01 pm

This change should map `/:queueName/paused` to the `queueJobsByState` instead of the `jobDetails` handler.

### DIFF
--- a/src/server/views/dashboard/index.js
+++ b/src/server/views/dashboard/index.js
@@ -7,7 +7,7 @@ const jobDetails = require('./jobDetails');
 
 router.get('/', queueList);
 router.get('/:queueHost/:queueName', queueDetails);
-router.get('/:queueHost/:queueName/:state(waiting|active|completed|succeeded|failed|delayed)\.:ext?', queueJobsByState);
+router.get('/:queueHost/:queueName/:state(waiting|active|completed|succeeded|failed|delayed|paused)\.:ext?', queueJobsByState);
 router.get('/:queueHost/:queueName/:id', jobDetails);
 
 module.exports = router;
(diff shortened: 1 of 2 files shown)


mshwery:
@ttacon any thoughts about this?

mshwery:
I think there a few other things we'll have to do to truly support this. Some of the data access for paused doesn't quite line up with the other views it seems.

mshwery:
I'll revisit this when I have some time for all those inner workings. :)

fbukevin:
Does this PR seem have not been merge yet?

fbukevin:
I tried with this modification in bull-arena, then I got error :
```
UnhandledPromiseRejectionWarning: TypeError: queue[(("get" + (intermediate value)) + "")] is not a function
```

bull-arena cannot map `getPaused` to function in bull. When I check the source in bull:
```js
  // TO BE DEPRECATED
  Queue.prototype.getPausedCount = function() {
    return this.getJobCountByTypes('paused');
  };
```

😂

PR #137, opened 2018-11-27
added yarn.lock and bumped up the nodemon version to remove flatmap-s…

SarasArya:
…tream as dependency

This is in response to [flatmap-stream package being deleted](https://github.com/dominictarr/event-stream/issues/116#issue-382854428). I upgraded the nodemon which in turns upgrades the ps.remy to 1.1.2


skeggse:
Duplicate of #184. Not considering inclusion of `yarn.lock` as we use npm's lockfile.

skeggse:
Released in v2.8.2.

PR #139, opened 2018-11-28
chore(dockerfile): use optimised alpine based image

### LINKED ISSUE #104: Use an optimized alpine docker image
The Dockerfile is based on node:8.1.2 (based on debian)

We could use the lighter image based on alpine, to run arena we don't need a big image like debian.

Also, we could be based on node:8-alpine, to get the last security updates automatically.

There is a number of optimizations that can be applied, based on node best practices 

https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md

I can make a PR if you want

sebaplaza:
Create alpine Dockerfile based on node best practices.

info: https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md

Closes #104

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM node:8-alpine
 
-# Tini for Handling Kernel Signals
-# https://github.com/krallin/tini
-# https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals
+# - Upgrade alpine packages to avoid possible os vulnerabilities
+# - Tini for Handling Kernel Signals https://github.com/krallin/tini
+#   https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals
 RUN apk --no-cache upgrade && apk add --no-cache tini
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,20 @@
...
+RUN apk --no-cache upgrade && apk add --no-cache tini
 
-RUN mkdir -p /opt/arena
 WORKDIR /opt/arena
-COPY package.json /opt/arena
-COPY package-lock.json /opt/arena
-RUN npm install --production
+
+COPY package.json package-lock.json /opt/arena/
+RUN npm ci --production && npm cache clean --force
```
skeggse: Why the `cache clean`? Is it to reduce disk util?

sebaplaza: yes, exactly, out image is lighter.

npm keeps a cache (in ~/.cache) of installed modules to speed up future installs (from other local projects).

In a docker image, we only need the /node_modules local folder, we' re not going to perform other installs.

In other words, we avoid to make a useless copy of our /node_modules folder.

PR #140, opened 2018-12-14
Bump handlerbars-helper to fix vulnerability

bkrausz:
https://nodesecurity.io/advisories/745 just came out, which affects versions of underscore.string prior to 3.3.5. bull-arena includes this via handlebars-helpers, which has had a more recent version out for a while now that uses a patched dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "express": "^4.15.2",
     "express-handlebars": "^3.0.0",
     "handlebars": "^4.0.10",
-    "handlebars-helpers": "^0.8.2",
+    "handlebars-helpers": "^0.10.0",
     "lodash": "^4.17.4",
     "nodemon": "^1.18.7"
   },


eeVoskos:
Hi @skeggse. Could you merge this and publish a new version on npm? Thanks.

skeggse:
I can, yes. It seems low priority to me because there's no way to actually exploit this against the arena - it's only used by `argparse` dependency, which is only used by the tests/binaries in `remarkable`.

skeggse:
@bkrausz would you please update the `package-lock.json` along with this change?

eeVoskos:
> I can, yes. It seems low priority to me because there's no way to actually exploit this against the arena - it's only used by `argparse` dependency, which is only used by the tests/binaries in `remarkable`.

@skeggse well, the only problem - for me at least :-/ - is that Google Cloud has become rather sensitive lately and is not allowing deployment of apps with npm audit warnings on AppEngine.

eeVoskos:
Your move @bkrausz !

skeggse:
Superseded by #151.

PR #146, opened 2019-01-31
Fixing navbar toggle in the main layout

anurag-rai:
On small screens (e.g. mobile devices) the `navbar` at the top of the screen does **not** function due to the missing reference to the `div`.


ttacon:
Awesome, thanks for the PR!

PR #55, opened 2017-09-28
Interface updates & Ability to add jobs

krazyjakee:

PR #153, opened 2019-02-25
Merge in job cleanup code

ttacon:
I pulled in #55 and cleaned up the merge conflicts also picking up a few changes that @skeggse and I had asked for. I'll merge this in and cut a release so that we can get this into folks hands.


ttacon:
Published as version 2.6.0

PR #155, opened 2019-02-27
PATCH: Commit vendor files and omit copying

ttacon:
Hot patch, will file follow on cleanup tasks for installation hooks.


ttacon:
Filed #156 for follow on work.

PR #157, opened 2019-02-27
Fix "add job" vendor/API path when basePath is set

jacobpgn:
The shiny new "add job" ✨codes need to follow the basePath prefix convention used elsewhere in order for everything to work correctly when a basePath is set – currently the JS/CSS assets won't be loaded, and the API request will 404!


ttacon:
Great catch! 🙌

ttacon:
Published as v2.6.2.

PR #123, opened 2018-09-26
Allow arena to work as a node module alongside auth middleware.

aaron-blondeau-dose:
I have been unable to find a way to add arena to an existing express app while also implementing authentication.  The basePath parameter conflicts with any path other than "/" provided to express' use method.  Allowing users to provide a new parameter called "mountPath" can alleviate this issue and allow configs like the following to work:

```
const arena = Arena({
    queues
}, {
    basePath: '/admin/',
    mountPath: '/',
    disableListen: true
})

app.use('/admin', basicAuth('foo', 'bar'), arena)
```

where basicAuth = https://github.com/expressjs/basic-auth-connect

PR #164, opened 2019-04-03
fix when baseUrl and routeUrl is not the same

### LINKED ISSUE #162: Can't be access if not deploy app on root
`basePath: '/arena'`
`app.use('/',arena)`
app deployed at example.com/app
arena external path: example.com/app/arena/

Diluka:
Fix problems that cannot be accessed when ``and`` are not the same

| scenario | `listenOpts.basePath` | `app.locals.appBasePath` |
| --- | --- | --- |
| standalone | use to mount route | keep empty |
| middleware | external base path | the same |

usage example
```ts
const url = new URL(config.get('server.url'));
app.use(
  '/arena',
  basicAuth({ users: { user: 'pass' }, challenge: true }),
  Arena(
    {
      queues: [
        {
          name: 'default',
          hostId: '默认',
          redis: config.get('redis[2]'),
        },
      ],
    },
    {
      disableListen: true,
      basePath: url.pathname,
    },
  ),
);
```

fix #162

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,10 +10,14 @@ function run(config, listenOpts = {}) {
   if (config) Queues.setConfig(config);
   Queues.useCdn = typeof listenOpts.useCdn !== 'undefined' ? listenOpts.useCdn : true;
 
-  app.locals.appBasePath = listenOpts.basePath || app.locals.appBasePath;
-
-  app.use('/', express.static(path.join(__dirname, 'public')));
-  app.use('/', routes);
+  if (listenOpts.disableListen) { 
+    app.locals.appBasePath = listenOpts.basePath || app.locals.appBasePath;
+    app.use('/', express.static(path.join(__dirname, 'public')));
+    app.use('/', routes);
+  } else {
+    app.use(listenOpts.basePath || app.locals.appBasePath, express.static(path.join(__dirname, 'public')));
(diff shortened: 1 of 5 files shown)


softbrix:
I like this configuration. Works as I expected. Downloaded this branch as a temporary fix until this is merged and deployed to npm.

softbrix:
Small update after som more testing. 
I had to configure the basePath like this for it to work when the application was hosted on the root path:
`basePath: baseUrlPath != '/' ? baseUrlPath : undefined,`

Dunno but I think this should be handled by the configuration so I don't need to set the basepath to 'undefined' since '/' is the default behaviour.

skeggse:
https://github.com/bee-queue/arena/pull/123#pullrequestreview-461292985:

>This has been a long-requested feature. I'd like to move this along, but still don't have a ton of time to review. I don't immediately understand the distinction between the `appBasePath` and the `mountPath`. Along those lines, could we get some documentation in the readme describing how the new option functions and how it differs from the old?

JoJ123:
@Diluka @softbrix @skeggse We are facing the same issue, can't use arena anymore with AppGateway. Is there a change that this will be released soon? This UI has some benefits vs other Bull UIs in the market.

PR #166, opened 2019-04-11
Add "Execute at" in template for delayed jobs.

### LINKED ISSUE #145: Show timestamp + delay info
Hi,

I'm not sure if I'm missing something or it's just not implemented, but for delayed jobs I can't seem to find the time of the actual execution of the job. Example: job is sent the 19th of January at 09:00am with a delay of 5 hours, but I can't see anywhere on the dashboard that the job will be executed the same day at 02:00pm.

I would be happy to shoot a PR for this, please just confirm that it's missing.

Best,
Flavio

JensVanhooydonck:
This will show the time when a delayed task will be executed.

Only tested this with Bull. 

Closes #145

### DIFF
--- a/src/server/views/partials/dashboard/jobDetails.hbs
+++ b/src/server/views/partials/dashboard/jobDetails.hbs
@@ -23,23 +23,25 @@
     <h5>Timestamp</h5>
     {{#if this.options.timestamp}}
       {{moment this.options.timestamp "llll"}}
-    {{/if}}
-    {{#if this.timestamp}}
+    {{else if this.timestamp}}
       {{moment this.timestamp "llll"}}
+    {{else}}
+      Malformed job.
     {{/if}}
   </div>
   
-  <div class="col-sm-3">
-    {{#if this.delay}}
-    <h5>Execute at</h5>
-    {{#if this.options.timestamp}}
-    {{moment (plus this.options.timestamp this.delay) "llll"}}
-    {{/if}}
-    {{#if this.timestamp}}
-    {{moment (plus this.timestamp this.delay) "llll"}}
-    {{/if}}
-    {{/if}}
-  </div>
+  {{#if this.delay}}
+    <div class="col-sm-3">
+      <h5>Execute at</h5>
+      {{#if this.options.timestamp}}
+        {{moment (plus this.options.timestamp this.delay) "llll"}}
+      {{else if this.timestamp}}
+        {{moment (plus this.timestamp this.delay) "llll"}}
+      {{else}}
+        Malformed job.
+      {{/if}}
+    </div>
+  {{/if}}
 
   <div class="col-sm-3">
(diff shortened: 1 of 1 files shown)


sibelius:
can we get this in?

sibelius:
screenshot of how it shows

image


you can use this patch-package for now https://gist.github.com/sibelius/a597174b1ff58c2883566693edb1fa15

until this lands

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -28,6 +28,18 @@
       {{moment this.timestamp "llll"}}
     {{/if}}
   </div>
+  
+  <div class="col-sm-3">
+    {{#if this.delay}}
```
skeggse, line 33: Will this work for bee-queue jobs? `this.delay` would be falsy, so it wouldn't ever enter this `if` clause - right?

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -28,6 +28,18 @@
...
+    <h5>Execute at</h5>
+    {{#if this.options.timestamp}}
+    {{moment (plus this.options.timestamp this.delay) "llll"}}
+    {{/if}}
+    {{#if this.timestamp}}
```
skeggse, line 38: If `this.delay` is truthy and `this.timestamp` is falsy, then there would just be a weird "Execute at" section with no content - right? Maybe this could be structured as:

```hbs
{{#if this.options.timestamp}}
...
{{else if this.timestamp}}
...
{{else}}
Malformed job.
{{/if}}
```

src/server/views/partials/dashboard/jobDetails.hbs
JensVanhooydonck: Right, I can only test this with Bull. I don't know how bee-queue will pass it's delay?
I could put the if this.delay around the div class, so it would not show any white space when there is no delay.

src/server/views/partials/dashboard/jobDetails.hbs
JensVanhooydonck: I have altered this piece. I added the same logic to the timestamp part so the code is consistent.

skeggse:
We don't have the bandwidth to test these changes against `bee-queue`, unfortunately. Happy to pick this back up if/when someone is able to verify this works for both bee-queue and bull!

roggervalf:
@JensVanhooydonck could you try last version #340 was merged

bradvogel:
I believe this was added in https://github.com/bee-queue/arena/pull/340

JensVanhooydonck:
> @JensVanhooydonck could you try last version #340 was merged

Sorry for the late reply. 
Everything works as expected. Thanks for this!

PR #169, opened 2019-05-01
Support pagination in JSON dump

bdentino:
This PR does not change the default behavior linked to on the UI, which is to dump the first 1000 jobs. It simply adds support for pagination to the endpoint so that you can request more than just the first 1000 jobs via JSON.

### DIFF
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -34,12 +34,27 @@ async function _json(req, res) {
 
   if (!isValidState(state, queue.IS_BEE)) return res.status(400).json({ message: `Invalid state requested: ${state}` });
 
+  const page = parseInt(req.query.page, 10) || 1;
+  const pageSize = parseInt(req.query.pageSize, 10) || 1000;
+
+  const startId = (page - 1) * pageSize;
+  const endId = startId + pageSize - 1;
+
   let jobs;
   if (queue.IS_BEE) {
-    jobs = await queue.getJobs(state, { size: 1000 });
+    const page = {};
+
+    if (['failed', 'succeeded'].includes(state)) {
+      page.size = pageSize;
+    } else {
+      page.start = startId;
+      page.end = endId;
+    }
+
+    jobs = await queue.getJobs(state, page);
     jobs = jobs.map((j) => _.pick(j, 'id', 'progress', 'data', 'options', 'status'));
   } else {
-    jobs = await queue[`get${_.capitalize(state)}`](0, 1000);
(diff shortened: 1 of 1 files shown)


src/server/views/dashboard/queueJobsByState.js
```diff
@@ -34,12 +34,27 @@ async function _json(req, res) {
 
   if (!isValidState(state, queue.IS_BEE)) return res.status(400).json({ message: `Invalid state requested: ${state}` });
 
+  const page = parseInt(req.query.page, 10) || 1;
+  const pageSize = parseInt(req.query.pageSize, 10) || 1000;
```
skeggse, line 38: We should probably enforce bounds on this, yeah?

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -34,12 +34,27 @@ async function _json(req, res) {
...
+  const startId = (page - 1) * pageSize;
+  const endId = startId + pageSize - 1;
+
   let jobs;
   if (queue.IS_BEE) {
-    jobs = await queue.getJobs(state, { size: 1000 });
+    const page = {};
+
+    if (['failed', 'succeeded'].includes(state)) {
+      page.size = pageSize;
```
skeggse, line 48: In this branch, should we return an error if `page` is specified? Otherwise, it might look like it's working but giving weird results.

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -34,12 +34,27 @@ async function _json(req, res) {
 
   if (!isValidState(state, queue.IS_BEE)) return res.status(400).json({ message: `Invalid state requested: ${state}` });
 
+  const page = parseInt(req.query.page, 10) || 1;
```
skeggse, line 37: What do you think about making this zero-based?

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -34,12 +34,27 @@ async function _json(req, res) {
...
   if (queue.IS_BEE) {
-    jobs = await queue.getJobs(state, { size: 1000 });
+    const page = {};
+
+    if (['failed', 'succeeded'].includes(state)) {
+      page.size = pageSize;
+    } else {
+      page.start = startId;
+      page.end = endId;
+    }
```
skeggse, line 52: This would be a bit terser, and maybe preferable.
```suggestion
    const page = ['failed', 'succeeded'].includes(state) ? { size: pageSize } : { start: startId, end: endId };
```

skeggse:
Closing due to inactivity. Feel free to reopen and revisit!

PR #179, opened 2019-08-13
Improvement/adding jobs

alexandrugheorghe:
Added support for named jobs and fixed json editor default content error.


skeggse:
mostly addressed by #209.

> fixed json editor default content error.

can you explain more about this?

alexandrugheorghe:
@skeggse I was referring to this fix: https://github.com/bee-queue/arena/pull/179/files#diff-7f05ff769b20fe816027a3f3240365c8R136 which checks if `data` exists before setting the content.

PR #183, opened 2019-09-25
Omit circular json structures from JSON responses

ghmeier:
Prior to this change, retrieving json job details would always fail
with `TypeError: Converting circular structure to JSON`. After this
change, the json value loads correctly.


shils:
How did this happen though? It worked at some point, right?

PR #186, opened 2019-10-07
Maintenance: Updated non-secure dependencies

joelmertanen:
Updated dependencies with `npm audit fix` to reduce the number of non-secure dependencies.


ttacon:
This has been addressed in a different PR - thanks for opening this though!

PR #188, opened 2019-10-08
Added resume to queueList

nephew92:
Showing the queue JobTypes count from the queue list view

Kazam_screenshot_00089


pefgoncalves:
+1

src/server/views/dashboard/queueList.js
```diff
@@ -1,8 +1,12 @@
-function handler(req, res) {
+async function handler(req, res) {
```
skeggse, line 1: We don't currently configure error handling for async functions and express - how should we address that?

PR #193, opened 2019-11-20
Adds functionality to view bull job logs

anuragagarwal561994:


src/server/views/dashboard/queueJobsByState.js
```diff
@@ -40,6 +40,10 @@ async function _json(req, res) {
     jobs = jobs.map((j) => _.pick(j, 'id', 'progress', 'data', 'options', 'status'));
   } else {
     jobs = await queue[`get${_.capitalize(state)}`](0, 1000);
+    jobs.forEach(async (job) => {
```
skeggse, line 43: The `forEach` Array method won't resolve or handle promises, so the async here won't be awaited and thus will introduce race conditions and unhandled rejections.

skeggse:
Closing due to inactivity. Feel free to re-open and revisit!

PR #200, opened 2019-12-06
feat(queues): add retry all failed button

drouian-m:
Add a new button on queue detail screen to retry all failed jobs.

Capture d’écran 2019-12-06 à 14 53 24

### DIFF
--- a/public/dashboard.css
+++ b/public/dashboard.css
@@ -144,3 +144,7 @@ pre {
   overflow: hidden;
 }
 
+.panel-heading .btn {
+  margin-left: 5px;
+}
+
--- a/public/dashboard.css
+++ b/public/dashboard.css
@@ -144,7 +144,7 @@ pre {
   overflow: hidden;
 }
 
-.panel-heading .btn {
+.btn-retry {
   margin-left: 5px;
 }
 
--- a/public/dashboard.js
+++ b/public/dashboard.js
@@ -68,7 +68,7 @@ $(document).ready(() => {
 
     const { queueHost, queueName } = window.arenaInitialPayload;
 
-    const r = window.confirm(`Retry all failed jobs queue "${queueHost}/${queueName}"?`);
+    const r = window.confirm(`Retry all failed jobs in queue "${queueHost}/${queueName}"?`);
     if (r) {
       $.ajax({
         method: 'POST',
--- a/public/dashboard.js
+++ b/public/dashboard.js
@@ -66,8 +66,7 @@ $(document).ready(() => {
     e.preventDefault();
     $(this).prop('disabled', true);
 
-    const queueName = $(this).data('queue-name');
-    const queueHost = $(this).data('queue-host');
+    const { queueHost, queueName } = window.arenaInitialPayload;
 
     const r = window.confirm(`Retry all failed jobs queue "${queueHost}/${queueName}"?`);
     if (r) {
--- a/public/dashboard.js
+++ b/public/dashboard.js
@@ -62,6 +62,30 @@ $(document).ready(() => {
     $jobBulkCheckboxes.prop('checked', this.checked);
   });
 
+  $('.js-retry-all-failed').on('click', function(e) {
+    e.preventDefault();
+    $(this).prop('disabled', true);
+
+    const queueName = $(this).data('queue-name');
+    const queueHost = $(this).data('queue-host');
+
+    const r = window.confirm(`Retry all failed jobs queue "${queueHost}/${queueName}"?`);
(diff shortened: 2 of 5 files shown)


public/dashboard.css
```diff
@@ -144,3 +144,7 @@ pre {
   overflow: hidden;
 }
 
+.panel-heading .btn {
```
ttacon: Would you mind adding an extra class for this instead of adding extra styling to an existing class?

public/dashboard.js
```diff
@@ -62,6 +62,29 @@ $(document).ready(() => {
     $jobBulkCheckboxes.prop('checked', this.checked);
   });
 
+  $('.js-retry-all-failed').on('click', function(e) {
+    e.preventDefault();
+    $(this).prop('disabled', true);
+
+    const { queueHost, queueName } = window.arenaInitialPayload;
+
+    const r = window.confirm(`Retry all failed jobs queue "${queueHost}/${queueName}"?`);
```
ttacon:
```suggestion
    const r = window.confirm(`Retry all failed jobs in queue "${queueHost}/${queueName}"?`);
```

src/server/views/api/queueRetry.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Retry all queue failed jobs
+ * @param {*} req 
+ * @param {*} res 
+ */
+async function handler(req, res) {
+  const { queueName, queueHost } = req.params;
+
+  const {Queues} = req.app.locals;
```
ttacon:
```suggestion
  const { Queues } = req.app.locals;
```

For consistency with the destructuring operation above this one.

src/server/views/api/queueRetry.js
```diff
@@ -0,0 +1,35 @@
...
+  if (!queue) return res.status(404).send({error: 'queue not found'});
+
+  const promises = [];
+  let jobs;
+  if (queue.IS_BEE) {
+    jobs = await queue.getJobs('failed');
+  } else {
+    jobs = await queue.getFailed();
+  }
+  jobs.forEach(job => promises.push(job.retry()));
```
ttacon:
```suggestion
  jobs.forEach((job) => promises.push(job.retry()));
```

src/server/views/api/queueRetry.js
```diff
@@ -0,0 +1,35 @@
...
+
+  try {
+    await Promise.all(promises);
+    return res.sendStatus(200);
+  } catch (e) {
+    const body = {
+      error: 'queue error',
```
ttacon:
```suggestion
      error: 'Unexpected error while retrying all jobs',
```

drouian-m:
@ttacon i just commit your suggested changes

bvallelunga:
Any updates on merging this, I would love this feature.

src/server/views/api/queueRetry.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const { Queues } = req.app.locals;
+
+  const queue = await Queues.get(queueName, queueHost);
+  if (!queue) return res.status(404).send({error: 'queue not found'});
+
+  const promises = [];
+  let jobs;
+  if (queue.IS_BEE) {
+    jobs = await queue.getJobs('failed');
```
skeggse, line 17: This will only return up to 100 jobs by default, so it's a bit misleading to suggest that this will retry all jobs. Bee-queue does not provide functionality to let you query all jobs as that can result in an unbounded (realistically it's memory-bounded) time period where the redis server is unresponsive. Additionally, having this page through all pages could result in an (actually) unbounded operation that runs in perpetuity (e.g. if there's another actor adding new jobs or failing jobs and there's no limit).

I'd like to suggest that this isn't a feature we should be supporting in bee-queue. Since bull-queue already made its bed by supporting a `getFailed` operation, I don't mind adding this functionality there.

src/server/views/api/queueRetry.js
```diff
@@ -0,0 +1,35 @@
...
+  } else {
+    jobs = await queue.getFailed();
+  }
+  jobs.forEach((job) => promises.push(job.retry()));
```
skeggse, line 21: This seems way terser/more readable:
```suggestion
  const promises = jobs.map((job) => job.retry());
```

skeggse:
Closing due to inactivity. Apologies for the PR churn/friction that blocked this. Feel free to re-open and update - I think the major blocker that I see here is that this isn't really tractable under bee-queue, and perhaps not worth addressing there unless we see a specific desire for that feature.

PR #201, opened 2020-01-09
Changes for Displaying Job logs in Job Details

ganeshcse2991:
Display Job logs on the job details page. It will help to easily view the logs associated with the job while processing it.

_NOTE: Bull Queue version updated to the latest 3.12.1 since the older version does not support adding or getting logs for the Job._

image

### DIFF
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -96,7 +96,7 @@ async function _html(req, res) {
     jobs = jobs.filter((job) => job);
   } else {
     jobs = await queue[`get${_.capitalize(state)}`](startId, endId);
-    jobs.map(async (job) => {
+    await jobs.map(async (job) => {
       let logs = await queue.getJobLogs(job.id);
       job.logs = (logs.logs || "No Logs");
       return job;
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -96,6 +96,11 @@ async function _html(req, res) {
     jobs = jobs.filter((job) => job);
   } else {
     jobs = await queue[`get${_.capitalize(state)}`](startId, endId);
+    jobs.map(async (job) => {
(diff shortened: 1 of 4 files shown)


src/server/views/dashboard/queueJobsByState.js
```diff
@@ -96,6 +96,11 @@ async function _html(req, res) {
     jobs = jobs.filter((job) => job);
   } else {
     jobs = await queue[`get${_.capitalize(state)}`](startId, endId);
+    jobs.map(async (job) => {
```
ttacon: You'll need to await on the top level returned promises from the `map` call here, otherwise it's not guaranteed that the logs will be retrieved for each job.

ttacon:
Thanks for the PR! Only one comment and then this is probably good to go!

ganeshcse2991:
@ttacon  Addressed the review comment and made changes for it.

Changes includes:
1) Awaiting on top level for async map
2) Fixing a bug in job details when logs are shown undefined.

ttacon:
Awesome, thanks!

ttacon:
Released as `v2.7.0`

ganeshcse2991:
Thanks :)

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -96,6 +96,11 @@ async function _html(req, res) {
     jobs = jobs.filter((job) => job);
   } else {
     jobs = await queue[`get${_.capitalize(state)}`](startId, endId);
+    await jobs.map(async (job) => {
```
Subv, line 99: jobs.map returns an Array, shouldn't you use `await Promise.all(jobs.map(...))` here instead of `await jobs.map(...)`? As it is, you're doing the equivalent of `await Promise.resolve(array)` and thus not actually waiting for the mapped promises to finish

Subv: Could this be related to #204?

PR #205, opened 2020-02-21
Add config options to customize title and brand

erdivartanovich:
Add options to customize title and navbar brand.
I would love to have this options available so I can have dashboard with may own *title* and *navbar brand* instead of the default value *Arena*

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,8 @@ function run(config, listenOpts = {}) {
   if (config) Queues.setConfig(config);
   Queues.useCdn = typeof listenOpts.useCdn !== 'undefined' ? listenOpts.useCdn : true;
 
+  app.locals.title = config.title || 'Arena';
+  app.locals.brand = config.brand || 'Arena';
   app.locals.appBasePath = listenOpts.basePath || app.locals.appBasePath;
 
   app.use(app.locals.appBasePath, express.static(path.join(__dirname, 'public')));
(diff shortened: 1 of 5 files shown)


bvallelunga:
@skeggse do you think we can get this reviewed?

index.js
```diff
@@ -9,6 +9,8 @@ function run(config, listenOpts = {}) {
   if (config) Queues.setConfig(config);
   Queues.useCdn = typeof listenOpts.useCdn !== 'undefined' ? listenOpts.useCdn : true;
 
+  app.locals.title = config.title || 'Arena';
+  app.locals.brand = config.brand || 'Arena';
```
skeggse, line 13: This will override the `defaultConfig.title` and `defaultConfig.brand` values from `app.js` - no?

src/server/config/index.json
```diff
@@ -7,5 +9,12 @@
...
   ],
-  "queues": []
+  "queues": [
+    {
+      "name": "my_queue",
+      "port": 6381,
+      "host": "127.0.0.1",
+      "hostId": "AWS Server 2"
+    }
+  ]
```
skeggse, line 19: I'd like to avoid adding this, as it'll get shipped as default (+ in the docker image) and I'd rather not have this cause problems for some poor soul who had something else running on 6381 for some strange reason.

skeggse:
If I'm not mistaken, this is blocked on a fix to the behavior of how defaults get overridden (not super confident in my understanding of this PR). Feel free to reopen if that isn't the case; closing due to inactivity.

Issue #170, opened 2019-05-13
Feature request: Support for named process when adding jobs


jesperp:
The "Add job" feature is great but currently doesn't support adding a job to a specific named process in a queue. Not sure how the UI should be for this, maybe a simple optional input field for the named process name?

sylvainlap:
+1

loris:
+1

skeggse:
Released in v3.2.0.

PR #209, opened 2020-03-13
Added support for named jobs

ErikMartensson:
Adds support for named jobs with Bull queues.

Fixes https://github.com/bee-queue/arena/issues/170 without polluting the job data with special properties.

WOyeLRvTQA

### DIFF
--- a/src/server/queue/index.js
+++ b/src/server/queue/index.js
@@ -106,8 +106,9 @@ class Queues {
    *
    * @param {Object} queue A bee or bull queue class
    * @param {Object} data The data to be used within the job
+   * @param {String} name The name of the Bull job (optional)
    */
-  async set(queue, data) {
+  async set(queue, data, name) {
     if (queue.IS_BEE) {
       return queue.createJob(data).save();
     } else {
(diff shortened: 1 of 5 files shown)


src/server/queue/index.js
```diff
@@ -80,12 +80,13 @@ class Queues {
...
    * @param {Object} data The data to be used within the job
+   * @param {String} name The name of the Bull job (optional)
    */
-  async set(queue, data) {
+  async set(queue, data, name) {
     if (queue.IS_BEE) {
       return queue.createJob(data).save();
     } else {
-      return queue.add(data, {
+      return queue.add(name || null, data, {
```
skeggse, line 89: Based on the bull source code, it looks like this might cause problems by passing in a non-string. Am I interpreting that right?

https://github.com/OptimalBits/bull/blob/013c51942e559517c57a117c27a550a0fb583aa8/lib/queue.js#L700-L704

skeggse:
@ErikMartensson curious - do we also need to add job names when we retry jobs?

https://github.com/bee-queue/arena/blob/d65dc768cb091269ca6329e6565986dc063d5474/src/server/views/api/jobRetry.js#L18

ErikMartensson:
@skeggse Sorry for the late response, vacation and all...

I noticed that you already fixed the job retry part. Thanks! The code didn't look like that when I wrote my PR.

Thanks for merging, I hope it will be useful. ✌

PR #211, opened 2020-04-04
Run jobs on demand

bvallelunga:
We currently use [Agenda](https://github.com/agenda/agenda) paired with [AgendaDash](https://github.com/agenda/agendash). One of the features AgendaDash provides is an ability to run jobs at any time, not just when they are failed state. We consider this a core feature that any replacement should have. 

We are moving to Bull/Arena and would love this feature to be included.

### DIFF
--- a/src/server/queue/index.js
+++ b/src/server/queue/index.js
@@ -85,10 +85,16 @@ class Queues {
     if (queue.IS_BEE) {
       return queue.createJob(data).save();
     } else {
-      return queue.add(data, {
-        removeOnComplete: false,
-        removeOnFail: false
-      });
+      const args = [
+        data,
+        {
+          removeOnComplete: false,
+          removeOnFail: false
+        }
+      ];
+
+      if (data.name) args.unshift(data.name)
+      return queue.add.apply(queue, data);
     }
   }
 }
--- a/src/server/views/api/jobRetry.js
+++ b/src/server/views/api/jobRetry.js
@@ -12,7 +12,7 @@ async function handler(req, res) {
   try {
     const jobState = queue.IS_BEE ? job.status : await job.getState();
 
-    if(jobState === 'failed' && typeof job.retry === 'function') {
+    if (jobState === 'failed' && typeof job.retry === 'function') {
       await job.retry();
     } else {
       await Queues.set(queue, job);
--- a/src/server/views/api/jobRetry.js
+++ b/src/server/views/api/jobRetry.js
@@ -10,7 +10,14 @@ async function handler(req, res) {
   if (!job) return res.status(404).send({error: 'job not found'});
 
   try {
-    await job.retry();
+    const jobState = queue.IS_BEE ? job.status : await job.getState();
+
+    if(jobState === 'failed' && typeof job.retry === 'function') {
+      await job.retry();
+    } else {
+      await Queues.set(queue, job);
+    }
+
     return res.sendStatus(200);
   } catch (e) {
     const body = {
--- a/src/server/views/dashboard/jobDetails.js
+++ b/src/server/views/dashboard/jobDetails.js
@@ -19,14 +19,9 @@ async function handler(req, res) {
     return res.json(_.omit(job, 'domain', 'queue', '_events', '_eventsCount'));
   }
 
-  let jobState;
-  if (queue.IS_BEE) {
-    jobState = job.status;
-  } else {
-    jobState = await job.getState();
-    let logs = await queue.getJobLogs(job.id);
-    job.logs = (logs.logs || "No Logs");
-  }
+  const jobState = queue.IS_BEE ? job.status : await job.getState();
+  job.showRetryButton = !queue.IS_BEE || jobState === 'failed';
+  job.retryButtonText = jobState === 'failed' ? 'Retry' : 'Trigger';
 
   return res.render('dashboard/templates/jobDetails', {
     basePath,
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -103,6 +103,12 @@ async function _html(req, res) {
     })
   }
 
+  for (const job of jobs) {
+    const jobState = queue.IS_BEE ? job.status : await job.getState();
+    job.showRetryButton = !queue.IS_BEE || jobState == 'failed';
+    job.retryButtonText = jobState == 'failed' ? 'Retry' : 'Trigger';
+  }
+
   let pages = _.range(page - 6, page + 7)
     .filter((page) => page >= 1);
   while (pages.length < 12) {
--- a/src/server/views/partials/dashboard/jobDetails.hbs
+++ b/src/server/views/partials/dashboard/jobDetails.hbs
@@ -9,7 +9,7 @@
 
 {{#if showRetryButton}}
   <button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
-    {{retryButtonText}}
+    {{ retryButtonText }}
   </button>
 {{/if}}
 
--- a/src/server/views/partials/dashboard/jobDetails.hbs
+++ b/src/server/views/partials/dashboard/jobDetails.hbs
@@ -7,11 +7,12 @@
   Remove
 </button>
 
-{{#eq jobState 'failed'}}
+{{#if showRetryButton}}
   <button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
-    Retry
+    {{retryButtonText}}
   </button>
-{{/eq}}
+{{/if}}
+
 
 <div class="row">
   <div class="col-sm-3">


bvallelunga:
@skeggse @ttacon would love your review.

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,12 @@ async function handler(req, res) {
   if (!job) return res.status(404).send({error: 'job not found'});
 
   try {
-    await job.retry();
+    if(await job.getState() === "failed") {
```
skeggse, line 13: `getState` isn't a function provided by bee's `Job` interface.

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,12 @@ async function handler(req, res) {
...
+      await job.retry();
+    } else {
+      await queue.add(job.name, job.data)
```
skeggse, line 16: `Queue#add` isn't provided by bee's `Queue` interface.

bvallelunga:
> This isn't functionality that's supported by bee. I can't speak to whether it's supported by bull. If it is, happy to approve with a guard that disables the additional functionality.
> 
> Also, please try to match the existing code style when submitting PRs. We should probably adopt `prettier` in this project just to make this a non-issue.

Thank @skeggse for the review! I just added functionality so that it works with both Bee and Bull. I think I updated the code style to match, if I missed anything let me know.

src/server/queue/index.js
```diff
@@ -85,10 +85,16 @@ class Queues {
...
-      });
+      let args = [
+        data,
+        {
+          removeOnComplete: false,
+          removeOnFail: false
+        }
+      ];
+
+      if(data.name) args.unshift(data.name)
```
skeggse, line 96:
```suggestion
      if (data.name) args.unshift(data.name)
```

src/server/queue/index.js
```diff
@@ -85,10 +85,16 @@ class Queues {
     if (queue.IS_BEE) {
       return queue.createJob(data).save();
     } else {
-      return queue.add(data, {
-        removeOnComplete: false,
-        removeOnFail: false
-      });
+      let args = [
```
skeggse, line 88:
```suggestion
      const args = [
```

src/server/queue/index.js
```diff
@@ -85,10 +85,16 @@ class Queues {
...
+        data,
+        {
+          removeOnComplete: false,
+          removeOnFail: false
+        }
+      ];
+
+      if(data.name) args.unshift(data.name)
+      return queue.add(...args);
```
skeggse, line 97: bee-queue currently supports Node 4, so we can't use this syntax.

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,19 @@ async function handler(req, res) {
   if (!job) return res.status(404).send({error: 'job not found'});
 
   try {
-    await job.retry();
+    let jobState;
+    if (queue.IS_BEE) {
+      jobState = job.status;
+    } else {
+      jobState = await job.getState();
+    }
```
skeggse, line 18: truly just a suggestion - not a necessary change
```suggestion
    const jobState = queue.IS_BEE ? job.status : await job.getState();
```

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,19 @@ async function handler(req, res) {
...
+
+    if(jobState === "failed" && typeof job.retry === "function") {
```
skeggse, line 20:
```suggestion
    if (jobState === 'failed' && typeof job.retry === 'function') {
```

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -7,11 +7,9 @@
   Remove
 </button>
 
-{{#eq jobState 'failed'}}
-  <button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
-    Retry
-  </button>
-{{/eq}}
+<button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
```
skeggse, line 10: This button will be visible even if it's not available due to the job residing on a bee-queue server. I'd like to preserve the behavior that the button only appears for failed jobs in bee-queue.

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,19 @@ async function handler(req, res) {
...
+      await job.retry();
+    } else {
+      await Queues.set(queue, job);
```
skeggse, line 23: Not sure how Bull's `name` feature works, but does it belong as a parameter here too so it can get passed to `queue.add`?

src/server/views/partials/dashboard/jobDetails.hbs
bvallelunga: I am not that familiar with how handlebars and locals work, is there an easy way in the view to check for this. I am thinking the if statement should be `!queue.IS_BEE || queue.state == "failed"`

src/server/views/api/jobRetry.js
bvallelunga: It is already in the bull job object. In `Queues.set` we extract it before calling `queue.add`

src/server/views/partials/dashboard/jobDetails.hbs
skeggse: The recommended practice is that you introduce a new local like `showRetryButton` that computes the check in js and then you can just do `{{#if showRetryButton}}` without encoding complex logic in the template itself.

bvallelunga: Scratch that, I figured it out.

bvallelunga: > The recommended practice is that you introduce a new local like `showRetryButton` that computes the check in js and then you can just do `{{#if showRetryButton}}` without encoding complex logic in the template itself.

Just saw this. That makes sense, let me know if the way I did it works for you.

skeggse: Hey, no worries! I would strongly prefer a DRYer approach where we don't duplicate the markup thrice - the `showRetryButton` approach would achieve that.

bvallelunga: Awesome! Just made the changes.

src/server/views/dashboard/jobDetails.js
```diff
@@ -19,14 +19,9 @@ async function handler(req, res) {
...
-  if (queue.IS_BEE) {
-    jobState = job.status;
-  } else {
-    jobState = await job.getState();
-    let logs = await queue.getJobLogs(job.id);
-    job.logs = (logs.logs || "No Logs");
-  }
+  const jobState = queue.IS_BEE ? job.status : await job.getState();
+  job.showRetryButton = !queue.IS_BEE || jobState == 'failed';
+  job.retryButtonText = jobState == 'failed' ? 'Retry' : 'Trigger';
```
skeggse, line 24: strict equality for unsurprising behavior
```suggestion
  job.showRetryButton = !queue.IS_BEE || jobState === 'failed';
  job.retryButtonText = jobState === 'failed' ? 'Retry' : 'Trigger';
```

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -103,6 +103,13 @@ async function _html(req, res) {
     })
   }
 
+  await jobs.map(async (job) => {
+    const jobState = queue.IS_BEE ? job.status : await job.getState();
+    job.showRetryButton = !queue.IS_BEE || jobState == 'failed';
+    job.retryButtonText = jobState == 'failed' ? 'Retry' : 'Trigger';
```
skeggse, line 109:
```suggestion
    job.showRetryButton = !queue.IS_BEE || jobState === 'failed';
    job.retryButtonText = jobState === 'failed' ? 'Retry' : 'Trigger';
```

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -103,6 +103,13 @@ async function _html(req, res) {
     })
   }
 
+  await jobs.map(async (job) => {
```
skeggse, line 106: This will result in `await`ing an `Array`, which won't actually wait for the promises within the array to complete. This can result in unhandled rejections, and likely introduces a race condition or other incorrect behavior resulting in the button not appearing when it should.

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -103,6 +103,13 @@ async function _html(req, res) {
...
+    const jobState = queue.IS_BEE ? job.status : await job.getState();
+    job.showRetryButton = !queue.IS_BEE || jobState == 'failed';
+    job.retryButtonText = jobState == 'failed' ? 'Retry' : 'Trigger';
+    return job
```
skeggse, line 110: Not used
```suggestion
```

src/server/views/dashboard/queueJobsByState.js
```diff
@@ -103,6 +103,13 @@ async function _html(req, res) {
...
+  })
```
skeggse, line 111:
```suggestion
  });
```

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -7,11 +7,12 @@
   Remove
 </button>
 
-{{#eq jobState 'failed'}}
+{{#if showRetryButton}}
   <button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
-    Retry
+    {{ retryButtonText }}
```
skeggse, line 12: style consistency
```suggestion
    {{retryButtonText}}
```

bvallelunga:
> Almost there! I'm mostly concerned about one logical error.

Awesome!! Just put up an update!

src/server/views/api/jobRetry.js
```diff
@@ -10,7 +10,14 @@ async function handler(req, res) {
   if (!job) return res.status(404).send({error: 'job not found'});
 
   try {
-    await job.retry();
+    const jobState = queue.IS_BEE ? job.status : await job.getState();
+
+    if(jobState === 'failed' && typeof job.retry === 'function') {
```
skeggse, line 15:
```suggestion
    if (jobState === 'failed' && typeof job.retry === 'function') {
```

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -7,11 +7,12 @@
   Remove
 </button>
 
-{{#eq jobState 'failed'}}
+{{#if showRetryButton}}
   <button class="btn btn-success js-retry-job" data-queue-host="{{ queueHost }}" data-queue-name="{{ queueName }}" data-job-id="{{ this.id }}">
-    Retry
+    {{retryButtonText}}
```
skeggse, line 12:
```suggestion
    {{ retryButtonText }}
```

bvallelunga:
@skeggse thanks for merging! What is your guys typical timeline to release to NPM. I would love to make the switch from Agenda to Bull this week if it's possible.

skeggse:
Landed in `v2.8.0`.

bvallelunga:
Amazing! Super exicting

bvallelunga:
@skeggse I just put up another PR that fixes an issue with named processes (something the Bull community has struggled with for a bit).

PR #214, opened 2020-04-09
Fix on-demand jobs with named processes

bvallelunga:


bvallelunga:
@skeggse would you mind giving this a review. It is super quick!

src/server/queue/index.js
```diff
@@ -80,21 +80,25 @@ class Queues {
...
       const args = [
+        name,
         data,
-        {
+        Object.assign({
           removeOnComplete: false,
           removeOnFail: false
-        }
-      ];
+        }, opts)
```
skeggse, line 96: You might consider using
```js
{
  removeOnComplete: false,
  removeOnFail: false,
  ...opts,
}
```

src/server/queue/index.js
```diff
@@ -80,21 +80,25 @@ class Queues {
    *
    * @param {Object} queue A bee or bull queue class
    * @param {Object} data The data to be used within the job
+   * @param {String} name The name of the process to be used (Bull only)
+   * @param {Object} opts Options to pass to the job (Bull only)
```
skeggse, line 84: Just to clarify, is the recommendation then that these be elided for bee-queue? I think I'd prefer to keep this interface consistent between the two - overloaded interfaces tend to become messy. Maybe we could always accept an options object that contains the name, and the options object could be used in bee-queue as well (separate PR)?

src/server/queue/index.js
```diff
@@ -80,21 +80,25 @@ class Queues {
...
           removeOnFail: false
-        }
-      ];
+        }, opts)
+      ].filter(function(arg) {
+        return arg !== null && arg !== undefined
+      })
 
-      if (data.name) args.unshift(data.name)
-      return queue.add.apply(queue, data);
```
bvallelunga, line 97: Looks like we missed this in the last PR, but it should have been `queue.add.apply(queue, args)` not `data`.

src/server/queue/index.js
```diff
@@ -80,21 +80,25 @@ class Queues {
...
-        {
+        Object.assign({
           removeOnComplete: false,
           removeOnFail: false
-        }
-      ];
+        }, opts)
+      ].filter(function(arg) {
+        return arg !== null && arg !== undefined
+      })
```
skeggse, line 99:
```suggestion
      ].filter((arg) => arg !== null && arg !== undefined);
```

src/server/queue/index.js
bvallelunga: Is that supported with the Node versions we have to support?

src/server/queue/index.js
bvallelunga: I assumed that Bee does not use named process or opts, but if it does I'm happy to add them the Bee createJob call.

bvallelunga: After taking a look at Bee documentation, I believe you are right on implementation. I have adjusted the code to accomediate

src/server/queue/index.js
skeggse: \*grumbles* probably not :(

src/server/queue/index.js
skeggse: bee-queue definitely doesn't support named jobs, but it does support options. I wouldn't advise adding options for bee-queue yet, but I'm hoping to make this forward-compatible with that change.

bvallelunga: Sounds good. I added support so that we could in the future.

src/server/views/api/jobRetry.js
```diff
@@ -15,11 +15,17 @@ async function handler(req, res) {
...
+      if (job.opts !== undefined && job.opts !== null) {
+        job.opts.repeat = undefined
+      }
+
       await Queues.set(queue, job);
     }
 
     return res.sendStatus(200);
   } catch (e) {
+    console.log(e)
```
skeggse:
```suggestion
```

src/server/views/api/jobRetry.js
```diff
@@ -15,11 +15,17 @@ async function handler(req, res) {
     if (jobState === 'failed' && typeof job.retry === 'function') {
       await job.retry();
     } else {
+      // On-demand jobs should not have repeat settings
+      if (job.opts !== undefined && job.opts !== null) {
+        job.opts.repeat = undefined
```
skeggse, line 20:
```suggestion
        job.opts.repeat = undefined;
```

src/server/views/api/jobRetry.js
```diff
@@ -15,6 +15,11 @@ async function handler(req, res) {
     if (jobState === 'failed' && typeof job.retry === 'function') {
       await job.retry();
     } else {
+      // On-demand jobs should not have repeat settings
+      if (job.opts !== undefined && job.opts !== null) {
```
skeggse, line 19:
```suggestion
      if (job.opts) {
```

src/server/queue/index.js
```diff
@@ -86,15 +86,16 @@ class Queues {
...
+        data.name,
+        data.data,
         {
           removeOnComplete: false,
-          removeOnFail: false
+          removeOnFail: false,
+          ...data.opts
         }
-      ];
+      ].filter((arg) => arg !== null && arg !== undefined);
```
skeggse, line 96: This change seems like it's altering the contract provided by `Queue.set`. How does that impact the rest of the codebase and the UI?

src/server/queue/index.js
bvallelunga: Okay I'll revert it back to `Object.assign`

src/server/queue/index.js
bvallelunga: Let me investigate.

bvallelunga: Yea this change would definitely change the contract. I think the only other option I can think without making a breaking changes version is to add an additional argument that has options in it.

```
async set(queue, data, options) {
    options = {
          name: String,
          opts: Object
     }
}
```

bvallelunga: I do think it is worth noting that I think the way it is in the current PR is actually a way better way. I have seen a bunch of PRs up that add things like named processes when adding jobs manually through the editor. Now the editor would support named process and Bull opts.

skeggse:
> Yea this change would definitely change the contract.

I'm generally pretty against making breaking changes unless it's really important. That, coupled with my pretty significant lack of knowledge around how bull works, how arena works, and what this PR is attempting to accomplish puts me in a bad position to review this PR. Sorry I can't help here. If you can get a prior contributor to this bit of code (that iirc has changed a bunch of times - I'd like to stop reviewing code for this particular bit of logic 🙃) to sign off on this, them I'm ok with it.

PR #215, opened 2020-04-17
Send correct data to queue

pturley:
Setup:
```
"bull": "^3.13.0",
"bull-arena": "^2.8.1",
```

When retrying a completed job using the job data was being sent under the data key. This actually triggers the job with the correct data.

### DIFF
--- a/src/server/queue/index.js
+++ b/src/server/queue/index.js
@@ -94,7 +94,7 @@ class Queues {
       ];
 
       if (data.name) args.unshift(data.name)
-      return queue.add.apply(queue, args);
+      return queue.add(data.data, args);
     }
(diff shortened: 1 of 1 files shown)


pturley:
@skeggse The basic issue was that when I would click the retry job button the data for the job wasn't passed correctly into the new job that was created. Enough time has passed, so I dont have the example any more, but the issue was that I could not successfully retry jobs with data in the same format.

skeggse:
Got it. If you run into this again, feel free to re-open!

skeggse:
I also suspect this might have been fixed by #223.

pturley:
👍

PR #219, opened 2020-05-08
Fix queue details view showing stale Redis stats

### LINKED ISSUE #218: Queue details view shows stale Redis statistics
**Node version:** 10.19.0
**Arena version:** 2.8.1
**Bull version:** 3.13.0
**ioredis version**: 4.16.3 

---

The [queue details view](https://github.com/bee-queue/arena/blob/744120be371291c8ca781930c3096117238f159f/src/server/views/dashboard/queueDetails.js) appears to present stale data within the ["Redis Statistics" table](https://github.com/bee-queue/arena/blob/744120be371291c8ca781930c3096117238f159f/src/server/views/dashboard/templates/queueDetails.hbs#L29-L45).

Metrics shown are consistent throughout the life of a given Arena middleware instance and fail to update as additional clients connect and resources are used: `connected_clients`, `used_memory`, and other fields do not reflect the current state of Redis, but rather the state upon first loading.

This issue appears to stem from the reliance on an undocumented behaviour of the queue's Redis client, which is evidently no longer demonstrated given the current [ioredis](https://github.com/luin/ioredis) release.

It seems a helper assumes that issuing a Redis `INFO` query has the client's `serverInfo` object updated:

https://github.com/bee-queue/arena/blob/744120be371291c8ca781930c3096117238f159f/src/server/views/helpers/queueHelpers.js#L33-L35

Unfortunately such is not the case, and ioredis updates the `serverInfo` object only after initially entering a ready state, and [only provided `enableReadyCheck` is truthy](https://github.com/luin/ioredis/blob/f275bc24de3825f80415a69ff227a45251dd1a3b/lib/redis/event_handler.ts#L71).

Due to the inconsistency regarding this `serverInfo` field, I suggest the result of the `INFO` query be parsed directly to extract the necessary information and ensure up-to-date stats are presented.

Rua-Yuki:
This PR fixes #218.

Rather than relying on inconsistent behaviour which ioredis no longer appears to demonstrate, the resulting data from issuing an `INFO` command is parsed directly.

As an alternative to recreating the wheel with an info parser here, we could always rely on a library such as [`node-redis-info`](https://github.com/FGRibreau/node-redis-info) to have things handled.

### DIFF
--- a/src/server/views/helpers/queueHelpers.js
+++ b/src/server/views/helpers/queueHelpers.js
@@ -28,11 +28,49 @@ function formatBytes(num) {
   return (neg ? '-' : '') + numStr + ' ' + unit;
 }
 
+/**
+ * Parses a Redis server info string as returned by `INFO`, generating an
+ * output object populated with fields based upon key-value pairs read.
+ * 
+ * @param {string} rawServerInfo - Redis server info as returned from the
+ * `INFO` command.
+ * 
+ * @returns {object} Returns a dictionary with key-value pairs based upon
+ * the parsed server info. 
+ */
+function parseRedisServerInfo(rawServerInfo) {
+  // Return just an empty object if we've not been given a proper string.
+  if (typeof rawServerInfo !== 'string') {
+    return {};
+  }
+
+  const out = {};
+  const lines = rawServerInfo.split('\r\n');
+  for (const line of lines) {
+    // Skip section names and empty lines.
+    if (line.startsWith('#') || !line.length) {
+      continue;
+    }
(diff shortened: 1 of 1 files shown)


skeggse:
It'd be nice if `ioredis` exposed their existing parsing mechanism so we don't have to choose a reimplementation strategy :\

bradvogel:
@Rua-Yuki Mind reviewing https://github.com/bee-queue/arena/pull/459 that is similar to this implementation? If that looks OK, I'd like to merge it.

PR #230, opened 2020-08-04
feat: remove app execution, explicit dependencies for docker-arena migration

skeggse:
Much of this functionality is moving to `docker-arena` per https://github.com/bee-queue/docker-arena/pull/1.

See also https://github.com/bee-queue/arena/commit/05f1484c464fb76e6ed8f775ea475e20fe11285a.


src/server/app.js
```diff
@@ -20,7 +20,7 @@ module.exports = function () {
 
   const Queues = require('./queue');
 
-  const queues = new Queues(defaultConfig);
+  const queues = new Queues({ ...defaultConfig, ...config });
```
shils, line 23: what happens if `queues` is present in both `defaultConfig` and `config`?

skeggse: Then `config` will override `defaultConfig`'s `queues` value, which is desired.

PR #223, opened 2020-06-04
Fix job retry

roychri:
This is to fix issue #222

PR #235, opened 2020-08-05
fix(retry-job): include job name

### LINKED ISSUE #234: Named jobs are triggered with __default__ name
v2.8.2 (I haven't upgraded to 3.0+, yet) broke how "triggering" named jobs were saved: the `name` is ignore and the `__default__` one is used.

v2.8.1 and below worked fine.

skeggse:
#223 broke retries and delayed job triggering. @roychri does this seem right?

Fixes #234.


roychri:
This looks good to me

PR #238, opened 2020-08-07
Display stacktraces when job is in deplayed or completed state

### LINKED ISSUE #206: Display stacktrace for delayed jobs
Bull treats jobs that failed but are having more attempts to be made as delayed. In this case, it is good to display the stacktrace of those.

I would remove the check for failed status when displaying stacktraces: if job has stacktraces - lets just dispay it anyway.

https://github.com/bee-queue/arena/blob/master/src/server/views/partials/dashboard/jobDetails.hbs#L76

bogdan:
Fixes #206


bradvogel:
@skeggse this look ok to merge?

skeggse (approved):
Seems like it should be fine. I'm not familiar with how bull exposes stack traces, though.

Issue #241, opened 2020-09-04
Job name is taked from inside job data


kikar:
Hi, I have a queue where the data contains a name property, and that name is used in the job list instead of the job name.

Is there a reason for this?

You can see it happening [here](https://github.com/bee-queue/arena/blob/master/src/server/views/dashboard/templates/queueJobsByState.hbs#L99)

eeVoskos:
This behaviour is intentional, and was added in https://github.com/bee-queue/arena/pull/74.

Do you have a better suggestion that covers this use case?

kikar:
The behaviour is not documented, so now I have all funky names in the list, rather than just using the Bull name. This prevents people to use the `name` key in the job's data.

Bull has a job name already, why do you need to fetch it from the job?

eeVoskos:
Well, I believe this dates back to the time that bull didn't have a job `name` property, and we had to add one in the job's `data`.

Perhaps, a more appropriate change would be to give priority to using `job.name` in the UI, and only use `job.data.name` if the former is `undefined`.

So, try to replace this:

```
{{#if this.data.name}}
  <span style="margin-left: 8px">{{ this.data.name }}</span>
{{else if this.name}}
  <span style="margin-left: 8px">{{ this.name }}</span>
{{/if}}
```

with this:

```
{{#if this.name}}
  <span style="margin-left: 8px">{{ this.name }}</span>
{{else if this.data.name}}
  <span style="margin-left: 8px">{{ this.data.name }}</span>
{{/if}}
```

*Disclaimer:* I haven't tested the above code change, it's just a suggestion that may make the required change more backwards compatible.

kikar:
Thank you for the suggestion. I've updated the PR: https://github.com/bee-queue/arena/pull/242

eeVoskos:
So, I just checked this issue and PR. I have two comments:

1. Yes, Bull has an optional `name` field in `queue.add(name?: string, data: object, opts?: JobOpts)`. However, this actually refers to the job *type* (see docs [here](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueadd)), meaning that the core purpose of the `name` arg is the use of named processors (see doc [here](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueprocess)).
[...]

PR #242, opened 2020-09-04
#241 - Don't use name from job data

kikar:
Solves #241

### DIFF
--- a/src/server/views/dashboard/templates/queueJobsByState.hbs
+++ b/src/server/views/dashboard/templates/queueJobsByState.hbs
@@ -96,8 +96,8 @@
             <a role="button" data-toggle="collapse" href="#collapse{{hashIdAttr this}}">
               <h4 class="header-collapse">
                 <span class="label label-default">{{#if this.id}}{{ this.id }}{{else}}<em>missing id</em>{{/if}}</span>
-                {{#if this.data.__name}}
-                  <span style="margin-left: 8px">{{ this.data.__name }}</span>
+                {{#if this.data.arenaName}}
+                  <span style="margin-left: 8px">{{ this.data.arenaName }}</span>
                 {{else if this.name}}
                   <span style="margin-left: 8px">{{ this.name }}</span>
                 {{else if this.data.name}}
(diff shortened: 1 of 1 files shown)


testinaweb (approved):

eeVoskos (approved):
I approve of this change, but the documentation needs to be updated as well, to reflect this feature.

src/server/views/dashboard/templates/queueJobsByState.hbs
```diff
@@ -96,10 +96,12 @@
             <a role="button" data-toggle="collapse" href="#collapse{{hashIdAttr this}}">
               <h4 class="header-collapse">
                 <span class="label label-default">{{#if this.id}}{{ this.id }}{{else}}<em>missing id</em>{{/if}}</span>
-                {{#if this.data.name}}
-                  <span style="margin-left: 8px">{{ this.data.name }}</span>
+                {{#if this.data.__name}}
+                  <span style="margin-left: 8px">{{ this.data.__name }}</span>
```
skeggse, line 100: Where is this `__name` field defined? This strikes me as pretty difficult to discover, and as not a previously defined pattern.

eeVoskos: Well, the thing is, it's not defined. Not yet! However, this is a very true user need. The job name (named processor) is not enough, as it is the same for every job. Using `job.data.name` is not documented in `bull`. So, instead of using a non-documented `bull` field, let's add one (`job.data.__name`) that will be documented on `arena`.

@skeggse do you have any alternatives to suggest?

skeggse: `arenaName` or `displayName` both seem possible. I'd strongly prefer we don't rely on anything that's `_` or `__`-prefixed for this functionality - this isn't a good purpose to use such a reserved name.

eeVoskos: Absolutely! The name itself is not a problem. @kikar let's change the attribute name to `job.data.arenaName` to make it more explicit.

kikar:
All done @skeggse.
What do you think? Can it be merged now?

skeggse (approved):

skeggse:
This probably needs documentation but I have no objection to shipping the feature in the meantime.

PR #245, opened 2020-09-13
chore(doc): fix bull example, improve queue constructor missing error msg

vigzmv:


skeggse:
Thanks for this - it sounds like there's some friction in adopting v3 that inspired this PR. I landed some of these changes in 42c089bff37a1838f353f435ccfde7f858e829dd. I'm not sure that the update to the error message will significantly improve the adoption experience - it's not self-evident from that message how to enact the change, so perhaps instead we should direct users to the README?

PR #250, opened 2020-10-09
Security fix: Upgrade Bootstrap to v3.4.1

Piccirello:
This fixes several vulnerabilities[0] that are present in the currently used Bootstrap v3.7.1.

[0] https://snyk.io/test/npm/bootstrap/3.3.7

PR #249, opened 2020-10-09
Security fix: Upgrade jQuery to v3.5.1

Piccirello:
This fixes several vulnerabilities[0] that are present in the currently used jQuery v3.2.1.

[0] https://snyk.io/test/npm/jquery/3.2.1

PR #253, opened 2020-10-20
Security fix: upgrade jQuery and Bootstrap

Piccirello:
Follow up of #249 and #250.


skeggse (approved):

Piccirello:
@skeggse Do you have a rough idea as to when this change will be shipped?

skeggse:
The automated release seems to be broken, and I don't have access to the account that has publish access. Haven't heard back from the person who has access to check.

PR #294, opened 2020-11-25
Support redis connection options in BullMQ

HalfdanJ:
BullMQ takes redis connections options in the format `{ connection: { host: ... }  }` instead of `{ redis: { host: ... } }`. 

This works well with connection settings as strings, havent tested passing in ioredis object


skeggse:
cc @gtpan77 thoughts? Should we elide the `redis` key in this case as well, given that it's unused?

gtpan77:
Yes, we should do this change. Thanks @HalfdanJ

skeggse (approved):

PR #324, opened 2021-02-17
Change add job text depending on hide prop on dashboard

roggervalf:
When clicking on add job, would be nice to change the message of that button, so people would know that by clicking again, the editor will be hidden. This PR tries to handle this.

### DIFF
--- a/public/dashboard.js
+++ b/public/dashboard.js
@@ -148,7 +148,11 @@ $(document).ready(() => {
   });
 
   $('.js-toggle-add-job-editor').on('click', function () {
-    $('.jsoneditorx').toggleClass('hide');
+    const addJobText = $('.js-toggle-add-job-editor').text();
+    const shouldHide = addJobText !== 'Add Job';
+    $('.jsoneditorx').toggleClass('hide', shouldHide);
+    $('.js-toggle-add-job-editor').text('Cancel');
+
     const job = localStorage.getItem('arena:savedJob');
     if (job) {
       const { name, data } = JSON.parse(job);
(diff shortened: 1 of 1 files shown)


bradvogel:
This doesn't seem like something that should be added to arena generically, right? Some folks might want to add jobs in quick succession without the refresh

roggervalf:
> This doesn't seem like something that should be added to arena generically, right? Some folks might want to add jobs in quick succession without the refresh

Thank you @bradvogel I didn't consider that case, I could remove that part, what do you think in the button message change depending on hide prop?

bradvogel:
Yup that looks fine

roggervalf:
Thank you, comments are addressed

public/dashboard.js
```diff
@@ -148,7 +148,12 @@ $(document).ready(() => {
   });
 
   $('.js-toggle-add-job-editor').on('click', function () {
-    $('.jsoneditorx').toggleClass('hide');
+    const addJobText = $('.js-toggle-add-job-editor').text();
+    const shouldHide = addJobText !== 'Add Job';
```
bradvogel, line 152: I can't get this to work. My repro steps:

1. Run the example queue from PR https://github.com/bee-queue/arena/pull/326
image
2. Click the "Add Job" button.

Expected result: the button changes to "Cancel"
Actual result: the button doesn't change:
image

bradvogel (changes requested):

public/dashboard.js
roggervalf: Thank you @bradvogel, I could try your example and fix this condition, please take a look :eyes:

bradvogel (approved):
Reviewed and tested.

PR #326, opened 2021-02-18
feat: add contributing guidelines and working example

bradvogel:
In reviewing PR https://github.com/bee-queue/arena/pull/324 I realized that there was no easy way to run the queue locally. This adds a basic working example and some simple documentation.

Issue #240, opened 2020-09-01
Order Jobs by id ASC/DESC


marcodali:
<img width="1413" alt="Screen Shot 2020-09-01 at 18 26 34" src="https://user-images.githubusercontent.com/70110654/91915833-2beb5200-ec81-11ea-965d-19a91447ec0d.png">

Can you please add the ability to order these job ids by ASC / DESC per page?
Thanks!

PR #327, opened 2021-02-19
Apply ordering for jobs when using Bull queue

roggervalf:
When showing the jobs list by state, would be nice to have the last created jobs at the top, this PR tries to handle that by applying ordering desc by default. Also I added a dropdown for ordering so a user could choose how to order jobs by id
#240 
tested with #326 

Selection_002
Selection_003
Selection_004
Selection_005
Selection_006


roggervalf:
Hi @bradvogel, could you please help me with some 👀 in this PR

bradvogel:
I don't have a way to test the Bull queue but I trust that you've tested this. The code looks good.

Separately, it'd be great to have another example (apart from the Bee Queue example in https://github.com/bee-queue/arena/pull/326) that uses Bull.

bradvogel (approved):

Issue #309, opened 2021-01-07
Action on bumping jsoneditor because of open XSS-vunerability


didrikg:
Hey,

Saw this open PR for bumping jsoneditor https://github.com/bee-queue/arena/pull/307

I just wanted to ask if there is any planned action on this as the current version has an open XSS-vunerability in it (see https://github.com/josdejong/jsoneditor/issues/1029).

Thank you!

roggervalf:
hey @didrikg, #334 was merged, so I think we are good to close this issue

PR #334, opened 2021-03-04
Fix retry bulk action

roggervalf:
Retry bulk button is throwing an error because the action is not the correct one as in delayed status is sending remove only as action, now it handles the retry action too. Also I upgraded jsoneditor as in ·#309 and still works as expected from example folder

Selection_007
Selection_008
Selection_009

### DIFF
--- a/example/bull.js
+++ b/example/bull.js
@@ -12,6 +12,23 @@ async function main() {
   const server = new RedisServer(REDIS_SERVER_PORT);
   await server.open();
 
+  const queue = new Bull('name_of_my_queue', {
+    redis: {
+      port: REDIS_SERVER_PORT,
+    },
+  });
+
+  // Fake process function to move newly created jobs in the UI through a few of the job states.
+  queue.process(async function (job) {
+    // Wait 5sec
+    await new Promise((res) => setTimeout(res, 5000));
+
+    // Randomly succeeds or fails the job to put some jobs in completed and some in failed.
+    if (Math.random() > 0.5) {
+      throw new Error('fake error');
+    }
+  });
+
   Arena(
     {
       Bull,
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
     "express": "^4.15.2",
     "express-handlebars": "^5.1.0",
     "handlebars": "^4.7.6",
-    "jsoneditor": "^5.9.6",
     "lodash": "^4.17.15",
     "moment": "^2.29.1",
     "tablesort": "^5.0.1"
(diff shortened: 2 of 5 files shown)


roggervalf:
@bradvogel could you please give some :eyes: on this PR when have some free time?

src/server/views/api/bulkAction.js
```diff
@@ -22,8 +22,10 @@ function bulkAction(action) {
       if (!_.isEmpty(jobs)) {
         const jobsPromises = jobs.map((id) => queue.getJob(decodeURIComponent(id)));
         const fetchedJobs = await Promise.all(jobsPromises);
-
-        const actionPromises = fetchedJobs.map((job) => job[action]());
+        const actionPromises =
+          action === 'retry'
+            ? fetchedJobs.map((job) => Queues.set(queue, job.data, job.name))
```
bradvogel, line 27: is this Bull specific?

bradvogel (changes requested):

package.json
```diff
@@ -9,7 +9,7 @@
     "express": "^4.15.2",
     "express-handlebars": "^5.1.0",
     "handlebars": "^4.7.6",
-    "jsoneditor": "^5.9.6",
+    "jsoneditor": "^9.2.0",
```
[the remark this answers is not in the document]
roggervalf: you are right, I couldn't find a reference, also trying the example without it and they are working. Tyty

src/server/views/api/bulkAction.js
roggervalf: I'm seeing that this logic is for both cases https://github.com/bee-queue/arena/blob/master/src/server/views/api/jobRetry.js#L18, but in case of bulkAction for retry action is only allowed for Bull https://github.com/bee-queue/arena/blob/master/src/server/views/dashboard/queueJobsByState.js#L126 in delayed state. What do you think about it?

bradvogel: When testing locally with Bee, I don't see jobs retrying. My repro steps:
1. Pull code from https://github.com/bee-queue/arena/pull/337 locally along with this PR
2. Run `npm run start:bee` from `/examples`
3. Create a few jobs in the UI:
image
4. After a few seconds, about half of those jobs will be failed. Select them and click Retry. Refresh the page to show the latest job state.
image

Expected result: those jobs are put back in the 'waiting' state and one is put in 'active'
Actual result: nothing happens. Those jobs stay in 'failed.

roggervalf: thank you @bradvogel, now it should work. Also I add same example in bull file to try

example/bull.js
```diff
@@ -12,6 +12,23 @@ async function main() {
   const server = new RedisServer(REDIS_SERVER_PORT);
   await server.open();
 
+  const queue = new Bull('name_of_my_queue', {
```
bradvogel, line 15: thanks!

bradvogel (approved):
tested locally and approved

Issue #220, opened 2020-05-13
Improve performance for bull queue when used as remote monitor


meotimdihia:
I tried to monitor remote Redis-server
I can't load the page if the queue number is more than 1000. I checked and saw `arena` was stopped in these lines:

https://github.com/bee-queue/arena/blob/master/src/server/views/dashboard/queueJobsByState.js#L106-L110
https://github.com/bee-queue/arena/blob/master/src/server/views/dashboard/queueJobsByState.js#L99-L103

And everything is fine if I use https://github.com/mixmaxhq/arena/blob/master/src/server/views/dashboard/queueJobsByState.js
He removed these lines and it works perfectly.

skeggse:
Feel free to submit a pull request to fix this!

roggervalf:
@meotimdihia could you please try the last version, there is one pr that was merged referencing this issue

meotimdihia:
Sorry, currently I don't use Bee or Bull queue for my project.

roggervalf:
> Sorry, currently I don't use Bee or Bull queue for my project.

No worries, could this issue be closed?

meotimdihia:
Yes, I close this issue. 😀

PR #341, opened 2021-03-11
Fix performance in job details

roggervalf:
In order to prevent a loading time when we have 1000 jobs in our tables, we could just show the logs when we click in job button for standalone page, this PR should solve #220

### DIFF
--- a/src/server/views/helpers/handlebars.js
+++ b/src/server/views/helpers/handlebars.js
@@ -18,6 +18,18 @@ const replacer = (key, value) => {
 // For jobs that don't have a valid ID, produce a random ID we can use in its place.
 const idMapping = new WeakMap();
 
+const getTimestamp = (job) => {
+  // Bull
+  if (job.timestamp) {
+    return job.timestamp;
+  }
+
+  // Bee
+  if (job.options && job.options.timestamp) {
+    return job.options.timestamp;
+  }
+};
+
 const helpers = {
   json(obj, pretty = false) {
     const args = [obj, replacer];
@@ -62,20 +74,18 @@ const helpers = {
   },
 
   getDelay(job) {
+    // Bull
     if (job.delay) {
-      return job.delay;
+      return job.delay + getTimestamp(job);
     }
 
-    return job.options ? job.options.delay : undefined;
-  },
-
-  getTimestamp(job) {
-    if (job.timestamp) {
-      return job.timestamp;
+    // Bee
+    if (job.options && job.options.delay) {
+      return job.options.delay + getTimestamp(job);
     }
-
-    return job.options ? job.options.timestamp : undefined;
   },
+
+  getTimestamp,
 
   encodeURI(url) {
     if (typeof url !== 'string') {
--- a/src/server/views/helpers/handlebars.js
+++ b/src/server/views/helpers/handlebars.js
@@ -61,8 +61,20 @@ const helpers = {
     return mapping;
   },
 
+  getDelay(job) {
+    if (job.delay) {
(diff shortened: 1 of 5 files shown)


roggervalf:
Please @bradvogel when you have some free time, give some :eyes: on this PR

src/server/views/partials/dashboard/jobDetails.hbs
```diff
@@ -45,15 +43,10 @@
...
+  {{#if (getDelay this)}}
   <div class="col-sm-3">
     <h5>Executes At</h5>
-    {{moment (add this.delay (getTimestamp this) ) "llll"}}
-  </div>
-  {{else if this.options.delay}}
-  <div class="col-sm-3">
-    <h5>Executes At</h5>
-    {{moment (add this.options.delay (getTimestamp this) ) "llll"}}
+    {{moment (add (getDelay this) (getTimestamp this) ) "llll"}}
```
bradvogel, line 49: minor thing: it'd be cleaner for getDelay to internalize the math to add the delay. It's a bit confusing to have logic implemented in the template.

src/server/views/helpers/handlebars.js
```diff
@@ -61,8 +61,20 @@ const helpers = {
...
+    if (job.delay) {
+      return job.delay;
+    }
+
+    return job.options ? job.options.delay : undefined;
+  },
+
   getTimestamp(job) {
-    return job.timestamp ? job.timestamp : job.options.timestamp;
+    if (job.timestamp) {
```
bradvogel, line 73: Small nitpick: This would be a slightly cleaner implementation:

```
  getTimestamp(job) {
    // Bull
    if (job.timestamp) {
      return job.timestamp;
    }

    // Bee
    if (job.options && job.options.timestamp) {
      return job.options.timestamp;
    }
  },
```

src/server/views/helpers/handlebars.js
```diff
@@ -61,8 +61,20 @@ const helpers = {
     return mapping;
   },
 
+  getDelay(job) {
+    if (job.delay) {
+      return job.delay;
+    }
+
+    return job.options ? job.options.delay : undefined;
```
bradvogel, line 69: See below - I would use getTimestamp() directly in this method in order to keep the business logic out of the template.

bradvogel (changes requested):
How do I see the logs locally? I tried running `npm run start:bull` in the examples folder, added a job, and let it complete. I opened the direct link to the job (http://localhost:4735/Queue%20Server%201/name_of_my_queue/3) and didn't see "Logs" in the UI

src/server/views/helpers/handlebars.js
roggervalf: sure, I could refactor it in my last commit tyty

roggervalf:
> How do I see the logs locally? I tried running `npm run start:bull` in the examples folder, added a job, and let it complete. I opened the direct link to the job (http://localhost:4735/Queue%20Server%201/name_of_my_queue/3) and didn't see "Logs" in the UI

I added a log example in bull example file, when getting into delayed job should appear it in the standalone page now

bradvogel (approved):

anuragagarwal561994:
Hello @bradvogel I guess instead of completely disabling this feature we can provide this as an option to the user.
Some of us don't have thousand jobs and more to that if the pagination settings are under control like on one page if we display only 10-20 jobs then it should not make such difference.

bradvogel:
Yes, agreed. I'll accept a PR to add an option

PR #353, opened 2021-03-31
Promote delayed job in bull

roggervalf:
This PR is for adding a button to promote delayed jobs for bull. Also I added an example for bullmq (note: this is just a simple example).
In order to test this new feature please:
1. Go into example directory.
2. run: npm install
3. run: npm run start:bullmq
4. Open arena in http://localhost:4735/Queue%20Server%201/name_of_my_queue

Selection_019

5. Go to delayed jobs
6. Click in the job row
7. Click in Promote button

Selection_020

8. You now should see that the job is in waiting 

Selection_021

PR #356, opened 2021-04-02
Adding customJsPath options

roggervalf:
This PR is for customizing the layout by providing a new feature: customJsPath, so it could be pass as a param. In order to test this PR:
1. Go to examples
2. Run: npm run start:bull
3. You should see a google link in nav bar at the button
Selection_022

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Arena({
 
   // Optionally include your own stylesheet
   customCssPath: 'https://example.com/custom-arena-styles.css',
+
+  // Optionally include your own script
+  customJsPath: 'https://example.com/custom-arena-js.js',
 });
 ```
 
@@ -194,6 +197,9 @@ const arenaConfig = Arena({
       },
     },
   ],
+
+  // Optionally include your own script
+  customJsPath: 'https://example.com/custom-arena-js.js',
 },
 {
   // Make the arena dashboard become available at {my-site.com}/arena.
@@ -238,6 +244,9 @@ const arenaConfig = Arena({
       },
     },
   ],
+
+  // Optionally include your own script
+  customJsPath: 'https://example.com/custom-arena-js.js',
 },
 {
   // Make the arena dashboard become available at {my-site.com}/arena.
--- a/example/bull.js
+++ b/example/bull.js
@@ -1,3 +1,5 @@
+const express = require('express');
+const path = require('path');
 const Arena = require('../');
 const Bull = require('bull');
 const RedisServer = require('redis-server');
@@ -33,7 +35,7 @@ async function main() {
   const delayedJob = await queue.add({}, { delay: 60 * 1000 });
   delayedJob.log('Log message');
 
-  Arena(
+  const app = Arena(
     {
       Bull,
 
@@ -54,11 +56,14 @@ async function main() {
           },
         },
       ],
+      customJsPath: 'example.js',
     },
     {
       port: HTTP_SERVER_PORT,
     }
   );
+
+  app.use(express.static(path.join(__dirname, 'public')));
 }
 
 main().catch((err) => {
--- a/example/bull.js
+++ b/example/bull.js
@@ -56,7 +56,7 @@ async function main() {
           },
         },
       ],
-      customJsPath: 'example.js',
+      customJsPath: 'http://localhost:4735/example.js',
     },
     {
       port: HTTP_SERVER_PORT,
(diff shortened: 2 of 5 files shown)


roggervalf:
@bradvogel please when you get a chance, some :eyes: in this PR

example/bull.js
```diff
@@ -57,6 +57,7 @@ async function main() {
     },
     {
       port: HTTP_SERVER_PORT,
+      externalLinks: [{ link: 'https://www.google.com/', text: 'google' }],
```
roggervalf, line 60: I added it for now, just to see in it with an example, after this PR is checked I could delete it

src/server/app.js
```diff
@@ -23,7 +23,6 @@ module.exports = function (config) {
   app.locals.Queues = queues;
   app.locals.appBasePath = '';
   app.locals.vendorPath = '/vendor';
-  app.locals.customCssPath = config.customCssPath;
```
roggervalf, line 26: From docs said that this should come from listenOpts param, not from config that is for queue configuration

README.md
```diff
@@ -167,6 +173,7 @@ router.use('/', arena);
 - `disableListen` - don't let the server listen (useful when mounting Arena as a sub-app of another Express app) (default: false)
 - `useCdn` - set false to use the bundled js and css files (default: true)
 - `customCssPath` - an URL to an external stylesheet (default: null)
+- `externalLinks` - an array of external links to be included in nav (default: null)
```
bradvogel, line 176: Can you share more about your use case for this? I'm concerned this is a "slippery slope" and there will be an explosion of different parameters to try to customize the UX. What we might need is some way to override the internal HTML templates that Arena uses.

roggervalf: sure, in my team we have some other projects related to queues that we would like to add in this nav for easy access when navigating between projects, could it be possible to have this in or something that I can change?

README.md
```diff
@@ -27,31 +27,37 @@ const Arena = require('bull-arena');
...
-  customCssPath: 'https://example.com/custom-arena-styles.css',
-});
+        // Queue key prefix. Defaults to "bq" for Bee and "bull" for Bull.
+        prefix: 'foo',
+      },
+    ],
+  },
+  {
+    // Optionally include your own stylesheet
+    customCssPath: 'https://example.com/custom-arena-styles.css',
```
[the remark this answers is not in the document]
roggervalf: I have a doubt here, in readme appears:
Selection_018
that customCssPath should be in listenOpts, not in queue config param :thinking: maybe also that part should be changed too :thinking:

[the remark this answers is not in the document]
bradvogel: Great catch. That was another bug in the readme. I fixed via a force push to https://github.com/bee-queue/arena/pull/359/files

README.md
bradvogel: What if we added a `customJsPath` parameter and you used it to write the links in using client-side JS? My thinking is that a `customJsPath` parameter would provide a lot more flexibility to customize anything.

README.md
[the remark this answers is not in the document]
roggervalf: nice let me revert these related changes

README.md
roggervalf: Sure let me change this

roggervalf:
in this PR I tried to handle the styled error with pr template https://github.com/bee-queue/arena/pull/353, this is why is failing

example/bull.js
```diff
@@ -54,11 +56,14 @@ async function main() {
           },
         },
       ],
+      customJsPath: 'example.js',
```
bradvogel, line 59: When I run the Bull example (via `npm run start:bull` from the examples dir) and navigate into a queue, I get the following 404 error:

image

README.md
```diff
@@ -51,6 +51,9 @@ Arena({
 
   // Optionally include your own stylesheet
   customCssPath: 'https://example.com/custom-arena-styles.css',
+
+  // Optionally include your own script
+  customJsPath: 'https://example.com/custom-arena-js.js',
```
bradvogel, line 56: looks like there's a merge conflict with master

bradvogel (changes requested):

example/bull.js
roggervalf: Good catch, I fixed it in last commit

README.md
roggervalf: I could fixed in last commits. Tyty

bradvogel (approved):

PR #360, opened 2021-04-06
Adding husky, pretty-quick and lint-staged

roggervalf:
#### Changes Made
This change I think will be very helpful because we are not going to fix the lint errors when developing. This could save time to people who contributes.
#### Potential Risks
To not fix lint errors manually xd

#### Test Plan
change your code and when you commit it, the file should be formatted automatically.

#### Checklist

- [ ] Since this is a public repository, I've checked I'm not publishing private data in the code, commit comments, or this PR.


roggervalf:
@bradvogel please when you get some time :eyes:

bradvogel:
What do you think about using the same eslint tooling and config from https://github.com/bee-queue/bee-queue (in this same organization)?

roggervalf:
> What do you think about using the same eslint tooling and config from https://github.com/bee-queue/bee-queue (in this same organization)?

sure I tried to use the same config but there were some inconsistencies between prettier and eslint rules that throws some errors pointing to bad lines, so for now I just use the same prettier as in bee-queue repo and use those as rules with eslint. What do you think?

bradvogel:
I'm a bit concerned about the auto-formatting part, as it's always a bit jarring to developers to have their code modified after committing. Most developers have auto-linting set up in their editors anyways that happens when the file is saved.

What's the advantage of doing this on commit vs what we currently do in that the CI blocks merging until it passes lint?

roggervalf:
> I'm a bit concerned about the auto-formatting part, as it's always a bit jarring to developers to have their code modified after committing. Most developers have auto-linting set up in their editors anyways that happens when the file is saved.
> 
> What's the advantage of doing this on commit vs what we currently do in that the CI blocks merging until it passes lint?

Yes, most people have the auto-linting, also I have it xd, but some developers don't have it, for example but using other editors different than Visual code, with the current ci, we just see those errors after we push the code on github, in this pr the errors will appear before you can commit your changes if the auto-formatting found an error that it can not be auto fixed. What do you think?

bradvogel (approved):
this sounds fine then

PR #389, opened 2021-05-20
Support waiting-children state for bullmq

roggervalf:
#### Changes Made
This PR adds a new state that is supported in bullmq: waiting-children, also is added the example into the example bullmq file.
#### Potential Risks
It's suppose to add a new bullmq state so should only be related to see in the example

#### Test Plan
Go to example directory and run: npm run start:bullmq
There are 2 queues, parentQueue should have 1 job in waiting-children state
If some of the children fails, we can retry those with the retry button
when those children are completed, the parent job in parentQueue should be processed

#### Checklist

### DIFF
--- a/src/server/views/dashboard/queueJobsByState.js
+++ b/src/server/views/dashboard/queueJobsByState.js
@@ -1,16 +1,30 @@
 const _ = require('lodash');
-const {BEE_STATES, BULL_STATES} = require('../helpers/queueHelpers');
+const {
+  BEE_STATES,
+  BULL_STATES,
+  BULLMQ_STATES,
+} = require('../helpers/queueHelpers');
 
+function getStates(isBee, isBullmq) {
+  if (isBee) {
+    return BEE_STATES;
+  }
+  if (isBullmq) {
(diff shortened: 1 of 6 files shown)


src/server/views/dashboard/queueJobsByState.js
```diff
@@ -32,7 +46,7 @@ async function _json(req, res) {
   const queue = await Queues.get(queueName, queueHost);
   if (!queue) return res.status(404).json({message: 'Queue not found'});
 
-  if (!isValidState(state, queue.IS_BEE))
+  if (!isValidState(state, queue.IS_BEE, queue.IS_BULLMQ))
```
bradvogel, line 49: What about just passing `queue` as the second parameter? It's more generic.

bradvogel (changes requested):

src/server/views/dashboard/queueJobsByState.js
roggervalf: sure, let me address it

bradvogel (approved):

PR #392, opened 2021-05-23
Feat flows view

roggervalf:
#### Changes Made
This PR adds new views to support flows creation (this is only available for Bullmq for now), so when the other queues allows the something similar, we could use the same pattern
#### Potential Risks
We would have the ability to create parent-children jobs in a single page view
#### Test Plan
1. Go to example directory
2. Run npm run start:bullmq
3. Go to Flows Overview
4. Enter to the first flow in the list
5. Click in the Add Flow button
6. Copy and paste this object:
```js
{
    "name": "parent-job2",
    "queueName": "name_of_my_parent_queue",
    "data": {},
    "children": [
      {"name": "child", "data": {"idx": 0, "foo": "bar"}, "queueName": "name_of_my_queue"},
      {"name": "child", "data": {"idx": 2, "foo": "bar"}, "queueName": "name_of_my_queue"},
{"name": "child", "data": {"idx": 3, "foo": "bar"}, "queueName": "name_of_my_queue"}
    ]
  }
```
7. All the jobs should be created, you now can go to queues overview and look for them, the parent should be in waiting-children state

Notes: 
1. Also I'm checking if the flows options is available, so bull and bee queues should not have the ability to see these views since this feature is not available for them yet.
2. As flows has a connection as the queues, it is following the same structure
3. Documentation for flows https://docs.bullmq.io/guide/jobs/flows

#### Checklist

### DIFF
--- a/example/bullmq.js
+++ b/example/bullmq.js
@@ -81,6 +81,8 @@ async function main() {
     {
       BullMQ: Queue,
 
+      FlowBullMQ: FlowProducer,
+
       queues: [
         {
           // Required for each queue definition.
@@ -103,6 +105,39 @@ async function main() {
 
           // User-readable display name for the host. Required.
           hostId: 'Queue Server 2',
+
+          // Queue type (Bull or Bullmq or Bee - default Bull).
+          type: 'bullmq',
+
+          redis: {
+            // host: 'localhost',
+            port: REDIS_SERVER_PORT,
+          },
+        },
+      ],
+
+      flows: [
+        {
+          // User-readable display name for the host. Required.
+          hostId: 'Flow server 1',
+
+          // Required for each flow definition.
+          name: 'Connection name 1',
+
+          // Queue type (Bull or Bullmq or Bee - default Bull).
+          type: 'bullmq',
+
+          redis: {
+            // host: 'localhost',
+            port: REDIS_SERVER_PORT,
+          },
+        },
+        {
+          // User-readable display name for the host. Required.
+          hostId: 'Flow server 2',
+
+          // Required for each flow definition.
+          name: 'Connection name 2',
 
           // Queue type (Bull or Bullmq or Bee - default Bull).
           type: 'bullmq',
--- a/src/server/flow/index.js
+++ b/src/server/flow/index.js
@@ -1,0 +1,131 @@
+const _ = require('lodash');
+
+class Flows {
+  constructor(config) {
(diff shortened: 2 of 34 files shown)


roggervalf:
@bradvogel please when you get a chance for some 👀 on this pr

src/server/flow/index.js
```diff
@@ -0,0 +1,131 @@
...
+      );
+    }
+  }
+
+  _checkConstructors() {
+    let hasBull = false,
+      hasBee = false,
+      hasBullMQ = false;
+    for (const flow of this._config.flows) {
+      if (flow.type === 'bee') hasBee = true;
```
bradvogel, line 47: When would this expression be true? AFAIK Bee doesn't have Flows, right? It would be nice to simplify the expression starting on line 54.

example/bullmq.js
```diff
@@ -113,6 +115,39 @@ async function main() {
           },
         },
       ],
+
+      flows: [
```
bradvogel, line 119: It might make sense to make a separate example for Flows, eg bullmq_with_flows.js

src/server/flow/index.js
```diff
@@ -0,0 +1,131 @@
+const _ = require('lodash');
+
+class Flows {
```
bradvogel, line 3: Add a comment that this is specific to BullMQ, unless I'm missing something and it works for other queues.

bradvogel (commented):
just a few small questions

src/server/flow/index.js
roggervalf: You are rigth, I tried to simplify it in last version

example/bullmq.js
roggervalf: Sure, I added that example in a separated file

src/server/flow/index.js
roggervalf: Sure, I added one in last version

roggervalf:
Please @bradvogel when you get a chance 👀

bradvogel:
Thanks. I'll review this by Tuesday at the very latest. We have a holiday weekend here

roggervalf:
> Thanks. I'll review this by Tuesday at the very latest. We have a holiday weekend here

No worries, when you get some free time please some 👀

bradvogel (approved):

bradvogel:
Apologies for the delay! Looks good.

PR #406, opened 2021-06-10
Use pagination with children

roggervalf:
#### Changes Made
This PR adds pagination for children, since getDependencies has options to do it. Note: Pagination is working with cursors so we could only go from one side, we cannot go back. This is the only limitation because  of how those dependencies are saved in Redis.
#### Potential Risks
No potential risk.

#### Test Plan
1. Go to example directory.
2. Run npm run start:bullmq
3. Go to waiting-children state in parent queue
4. Select the parent job
5. Go to the perma link to see job details
6. You should see a button for children to >> go to the next children jobs. there is a min count of children that is 25 to be showed.
#### Checklist


roggervalf:
Selection_037

bradvogel (approved):

PR #459, opened 2021-11-16
Fix queue details not loading - because of switch from node-redis to ioredis

oyvindneu:
I see you already have a PR for this #219 fixing issue #218 - but now you have another option.

Waiting for ioredis to implement this feature to fix this problem seems futile.

Looking at their code (below) it is not very advanced parsing only good enough to use the result when doing readyCheck. 

In order for them to expose this API they they need to write new robust code and it seems like this is not prioritized and it is a bigger task than one off parsing for use in Arena. The risk of doing this in Arena is smaller since - since Arena use a list of allowed fields anyway, so any parsing errors will be hidden from the en user.

https://github.com/luin/ioredis/blob/master/lib/redis/index.ts#L558

```
    const info: { [key: string]: any } = {};

    const lines = res.split("\r\n");
    for (let i = 0; i < lines.length; ++i) {
      const [fieldName, ...fieldValueParts] = lines[i].split(":");
      const fieldValue = fieldValueParts.join(":");
      if (fieldValue) {
        info[fieldName] = fieldValue;
      }
    }
```

### DIFF
--- a/src/server/views/helpers/queueHelpers.js
+++ b/src/server/views/helpers/queueHelpers.js
@@ -31,12 +31,35 @@ function formatBytes(num) {
   return (neg ? '-' : '') + numStr + ' ' + unit;
 }
 
+function splitInfo(res) {
+  if (typeof res !== 'string') {
+    return {};
+  }
+
+  const serverInfo = {};
+  const lines = res.split('\r\n');
+  for (let i = 0; i < lines.length; ++i) {
+    if (lines[i]) {
+      const line = lines[i].trim();
+      if (!line.startsWith('#')) {
+        const idx = line.indexOf(':');
+        if (idx > 0) {
+          serverInfo[line.substring(0, idx)] = line.substring(idx + 1);
+        }
+      }
+    }
+  }
+
+  return serverInfo;
+}
+
 const Helpers = {
   getStats: async function (queue) {
     const client = await queue.client;
-    await client.info(); // update queue.client.serverInfo
+    const info = await client.info(); // In node-redis this will update queue.client.serverInfo
 
-    const stats = _.pickBy(client.serverInfo, (value, key) =>
+    // In ioredis we need to parse this information:
+    const stats = _.pickBy(client.serverInfo || splitInfo(info), (value, key) =>
       _.includes(this._usefulMetrics, key)
     );
     stats.used_memory = formatBytes(parseInt(stats.used_memory, 10));


bradvogel:
Do you have an example that uses this feature?

oyvindneu:
Sorry for the late reply.

It fixes the regression with Redis statistics not being loaded after bullmq switched to ioredis.

At the moment it looks like this - note the "Could not retrieve value" - this is because arena can't read the result of client.info() the same way it could before (when bullmq relied on node-redis).

image

bradvogel:
@oyvindneu Thanks for the ping and apologies for not responding before. Mind adding a quick unit test for this function?

@Rua-Yuki Mind taking a look at this? It's similar to your PR https://github.com/bee-queue/arena/pull/219.

src/server/views/helpers/queueHelpers.js
```diff
@@ -31,12 +31,35 @@ function formatBytes(num) {
...
+
 const Helpers = {
   getStats: async function (queue) {
     const client = await queue.client;
-    await client.info(); // update queue.client.serverInfo
+    const info = await client.info(); // In node-redis this will update queue.client.serverInfo
 
-    const stats = _.pickBy(client.serverInfo, (value, key) =>
+    // In ioredis we need to parse this information:
+    const stats = _.pickBy(client.serverInfo || splitInfo(info), (value, key) =>
```
Rua-Yuki, line 62: I'd expect `client.serverInfo || splitInfo(info)` to produce erratic/unreliable behaviour.

For example, given `enableReadyCheck: true` ([the default for `ioredis` currently](https://github.com/luin/ioredis/blob/abd9a82433ad67b91a4bddb45aff8da2e20d45e8/API.md#new-redisport-host-options)) we will have `client.serverInfo` defined and thus used in this instance. **_This will result in the same stale data issue which #218 describes._**

Of course, given `enableReadyCheck: false` the `client.serverInfo` is undefined and thus we resort to `splitInfo(info)`. Only in this case will accurate and fresh data be presented.

It'd be better to parse each time so we've fresh data regardless the configuration of `enableReadyCheck`.

Rua-Yuki: As of Bull v4.0.0 we do in fact have `enableReadyCheck: false` as a guarantee on ioredis clients. That release is still relatively new however and it'd be preferable to solve the problem for older versions of Bull as well considering it is extremely simple to do so.

Rua-Yuki (changes requested):
**RE [comment 1027944237](https://github.com/bee-queue/arena/pull/459#issuecomment-1027944237) from @bradvogel on 02 Feb 2022 13:31 UTC:**
> @Rua-Yuki Mind taking a look at this? It's similar to your PR #219.

Sure thing!

This PR #459 solves #218 satisfactorily only in cases where `enableReadyCheck` is guaranteed to be falsy. This is only the case with Bull ≥ 4.0.0 (see https://github.com/OptimalBits/bull/issues/1873#issuecomment-952786853). 

For older versions of Bull, where we've ioredis' ready-check enabled by default, stale data is erratically presented still. Please see my review comments for more information.

This PR does technically solve an issue where queue details aren't presented at all in the absence of a ready check, though this is a different fundamental issue, and only a portion of #218:

> Unfortunately such is not the case, and ioredis updates the `serverInfo` object only after initially entering a ready state, and [only provided `enableReadyCheck` is truthy](https://github.com/luin/ioredis/blob/f275bc24de3825f80415a69ff227a45251dd1a3b/lib/redis/event_handler.ts#L71).

---

**RE [comment 995975417]() from @oyvindneu on 16 Dec 2021 16:26 UTC:**

> At the moment it looks like this - note the "Could not retrieve value" - this is because arena can't read the result of client.info() the same way it could before (when bullmq relied on node-redis).

Presentation of "Could not retrieve value" in particular is not the consequence of a deficiency in reading the result of `client.info()` but simply the entire absence of stored server info – almost definitely the result of `enableReadyCheck` being set falsy. (i.e., result is of no consequence but side-effects are)

This may seem like a pedantic distinction though I feel it is important to note because the described fix is off-target the referenced issue and there may be some confusion here.

PR #536, opened 2022-06-24
Fix queue details view #218

TKDBB84:
#### Changes Made
There [exists another PR Here](https://github.com/bee-queue/arena/pull/459) that addresses this same issue, however after feedback that PR seems to have been abandoned.  This is the same as above but I've addressed the feedback given, and am hoping to get a fix in for the stats display.  Please see the previous PR #459  and Issue #218 for context.


TKDBB84:
@roggervalf  - any interest in merging this?

PR #541, opened 2022-07-10
fix: BullMQ's delayed execution

SamerAlsayegh:
#### Changes Made
- Added handling for delayed execution dates using BullMQ.  (`job.opts.delay` is used in BullMQ vs  `job.delay` in Bull) 

Sample; 
image

#### Potential Risks
- Simple change, should only impact BullMQ jobs per the object structure

#### Test Plan
- Set up a delayed job using BullMQ
- Check it in Arena, and ensure the delayed date is not the current time and is the future intended date instead.

#### Checklist

- [ ] Since this is a public repository, I've checked I'm not publishing private data in the code, commit comments, or this PR.

### DIFF
--- a/src/server/views/helpers/handlebars.js
+++ b/src/server/views/helpers/handlebars.js
@@ -83,6 +83,11 @@ const helpers = {
     if (job.options && job.options.delay) {
       return job.options.delay;
     }
+
+    // BullMQ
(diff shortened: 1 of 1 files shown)


roggervalf:
hi @SamerAlsayegh, I think version 1.89.0 handles delay in job instance

PR #593, opened 2022-12-03
Security patches for qs dependency

elsilentforce:
#### Changes Made
`qs` before 6.10.3, as used in `Express` before 4.17.3 and other products, allows attackers to cause a Node process hang for an Express application because an __ proto__ key can be used. In many typical Express use cases, an unauthenticated remote attacker can place the attack payload in the query string of the URL that is used to visit the application, such as a[__proto__]=b&a[__proto__]&a[length]=100000000. The fix was backported to `qs` 6.9.7, 6.8.3, 6.7.3, 6.6.1, 6.5.3, 6.4.1, 6.3.3, and 6.2.4 (and therefore Express 4.17.3, which has "deps: qs@6.9.7" in its release description, is not vulnerable).
[Link to NVD](https://nvd.nist.gov/vuln/detail/cve-2022-24999)

This PR includes the following changes:

**body-parser**
Current version in use: `^1.17.2`
`qs` version used as sub dependency: [6.4.0](https://github.com/expressjs/body-parser/blob/1.17.2/package.json#L19)

Proposed change: `^1.20.0`
`qs` version used as sub dependency: [6.10.3](https://github.com/expressjs/body-parser/blob/1.20.0/package.json#L20)

**express**
Current version in use: `^4.15.5`
`body-parser` version used as sub dependency: [1.18.1](https://github.com/expressjs/express/blob/4.15.5/package.json#L61)

Proposed change: `^1.18.0`
`body-parser` version used as sub dependency: [1.20.0](https://github.com/expressjs/express/blob/4.18.0/package.json#L33)

After updating these packages and running a `npm ls qs` command, there is just references to safe versions:
Screenshot from 2022-12-03 12-06-21
_NOTE: @semantic-release still running under a unsafe version of qs but its used just as devDependency_

#### Potential Risks
If no minor version is released with this fix, all the users using this package could have some dependency issues

r#### Test Plan
Manual testing on another app with a big dependency ecosystem.

#### Checklist


package-lock.json
```diff
@@ -1,8 +1,10026 @@
 {
   "name": "bull-arena",
   "version": "3.30.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
roggervalf, line 4: could you use node 14 to generate package-lock changes as it was using version 1?

roggervalf (commented):
hi @elsilentforce thank you for this pr, just one comment

roggervalf:
also, as we are using semantic release, your commits should follow https://github.com/semantic-release/semantic-release rules, for example you could create a commit with this message:
`fix(qs): security patches on body-parser and express dependencies`

package-lock.json
elsilentforce: hello @roggervalf !
My bad, fixing it immediately.

roggervalf (approved):
LGTM

PR #623, opened 2023-02-17
fix(mounting): consider base and appBase paths

roggervalf:
#### Changes Made
Use basePath when disableListen is true. This may fix the case when using fastify
#### Checklist


richardmaysundo (approved):
LGTM

PR #662, opened 2023-08-21
feat(bullmq): support prioritized job state

patrickheeney:
#### Changes Made
Added additional [JobState](https://api.docs.bullmq.io/types/v4.JobState.html) recently introduced.
Reference: https://docs.bullmq.io/guide/jobs/prioritized
Reference: https://bullmq.io/news/062123/faster-priority-jobs/

#### Potential Risks
Tested on older bullmq and it ignores the prioritized state since it doesn't exist.

#### Test Plan
Screenshot 2023-08-20 at 8 31 45 PM

#### Checklist

- [X] I've increased test coverage
- [X] Since this is a public repository, I've checked I'm not publishing private data in the code, commit comments, or this PR.

### DIFF
--- a/src/server/views/dashboard/queueDetails.js
+++ b/src/server/views/dashboard/queueDetails.js
@@ -18,8 +18,6 @@ async function handler(req, res) {
   if (queue.IS_BEE) {
     jobCounts = await queue.checkHealth();
     delete jobCounts.newestJob;
-  } else if (queue.IS_BULLMQ) {
-    jobCounts = await queue.getJobCounts(...QueueHelpers.BULLMQ_STATES);
   } else {
     jobCounts = await queue.getJobCounts();
   }
--- a/src/server/views/helpers/queueHelpers.js
+++ b/src/server/views/helpers/queueHelpers.js
@@ -81,6 +81,7 @@ const Helpers = {
    */
   BULLMQ_STATES: [
     'waiting',
+    'prioritized',
     'active',
     'completed',
     'failed',


patrickheeney:
Created patch file for patch-package.

`patches/bull-arena+3.30.4.patch`
```
diff --git a/node_modules/bull-arena/src/server/views/helpers/queueHelpers.js b/node_modules/bull-arena/src/server/views/helpers/queueHelpers.js
index 2e0f2e0..c37b56f 100644
--- a/node_modules/bull-arena/src/server/views/helpers/queueHelpers.js
+++ b/node_modules/bull-arena/src/server/views/helpers/queueHelpers.js
@@ -87,6 +87,7 @@ const Helpers = {
     'delayed',
     'paused',
     'waiting-children',
+    'prioritized',
   ],
 };
 ```

File: src/server/views/helpers/queueHelpers.js [hunk 1]
```diff
@@ -87,6 +87,7 @@ const Helpers = {
     'delayed',
     'paused',
     'waiting-children',
+    'prioritized',
```
roggervalf, line 90: please move it after waiting state

roggervalf (commented):
as bullmq v3 does have this new state and if someone click on prioritized page from previous version, an error page will be shown. So or we treat this change as a breaking change or handle a way to check bullmq version to conditionally show this new state

patrickheeney:
> an error page will be shown. 

I am unable to replicate in my testing. Do you happen to have a screenshot of the error?

There shouldn't be any clickable link returned from `jobCounts` on older versions because [arena](https://github.com/bee-queue/arena/blob/master/src/server/views/dashboard/queueJobsByState.js#L99C29-L99C41) does not pass any arguments to `getJobCounts`, thus `bullmq` [uses its own internal list](https://github.com/taskforcesh/bullmq/blob/master/src/classes/queue-getters.ts#L137) instead of the ones defined by `arena` which wouldn't include the new one.


The prioritized state is ignored because no jobs match.

roggervalf:
> I am unable to replicate in my testing. Do you happen to have a screenshot of the error?
hey @patrickheeney sure
Captura desde 2023-08-24 23-23-38

roggervalf:
ok, I tried the last approach, but I prefer the ordering that was before and by adding prioritized state, I think it should appear in this order, people usually may look first for waiting states (waiting/prioritized), then active, completed and failed
```
'waiting' // first state that most jobs will start
'prioritized', // like waiting but with priority option, has less prevalence than waiting
'active', // after being in waiting or prioritized, it will be moved to active
'completed', // final state
'failed', // final state but with an error
'delayed', // it will be moved to waiting after some delay
'paused', // will not be moved until queue is resumed
'waiting-children' // if jobs are parents of other jobs
```
So for adding prioritized I may consider it as a breaking change in order to preserve this order and not confuse people using bullmq v3 with a broken page

patrickheeney:
Sorry, I am a little confused as to what you want me to do?  Doesn't the order in the latest commit, match what you are requesting? Also there shouldn't be a 404 because it isn't returned from `jobCounts` in order to click on it, unless you are specifically typing in that URL or I am missing something?

roggervalf:
hi @patrickheeney, I meant that the order is not respected with your last change, that was the purpose of BULLMQ_STATES to keep keep that order on our table

patrickheeney:
@roggervalf Sorry, I am still trying to understand.

You mentioned this about the order:
Screenshot 2023-08-31 at 7 46 04 PM

And this:
Screenshot 2023-08-31 at 7 46 21 PM

And this is the order of the last change:
Screenshot 2023-08-31 at 7 46 36 PM

Doesn't this match what you are suggesting? If not, can you please list them how you want it to be?

src/server/views/dashboard/queueDetails.js
```diff
@@ -18,8 +18,6 @@ async function handler(req, res) {
   if (queue.IS_BEE) {
     jobCounts = await queue.checkHealth();
     delete jobCounts.newestJob;
-  } else if (queue.IS_BULLMQ) {
```
roggervalf, line 21: yeah, but after removing these lines the order changed

roggervalf:
I'll fix it in another pr no worries

roggervalf (approved):

PR #668, opened 2023-11-14
add-favicon: implictly include favicon at layout template

ivnnv:
Just adding the favicon as its not showing

### DIFF
--- a/src/server/views/layout.hbs
+++ b/src/server/views/layout.hbs
@@ -22,6 +22,8 @@
     {{/if}}
     <link rel="stylesheet" href="{{ basePath }}/dashboard.css">
     <script>hljs.initHighlightingOnLoad();</script>
+    <!-- Favicon -->
+    <link rel="shortcut icon" href="{{ basePath }}/favicon.ico" type="image/x-icon">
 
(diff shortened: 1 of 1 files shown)


src/server/views/layout.hbs
```diff
@@ -22,6 +22,8 @@
     {{/if}}
     <link rel="stylesheet" href="{{ basePath }}/dashboard.css">
     <script>hljs.initHighlightingOnLoad();</script>
+    <!-- Favicon -->
+    <link rel="shortcut icon" href="{{ basePath }}/favicon.ico" type="image/x-icon">
```
roggervalf, line 26: what about passing it as a new option like customCssPath, also this should be optional, only if customFavicon is provided:
```
{{#if customFavicon}}
    <link rel="shortcut icon" href="{{ customFavicon }}" type="image/x-icon">
{{/if}}
```

ivnnv: Hi, I think I dont understand the comment, why having something like a custom favicon?
This PR intention is just showing the one that is already present at `/public/favicon.ico` but its not used.

roggervalf: hey @ivnnv I tried your change locally, lgtm thank you

roggervalf (approved):

